PR #1, opened 2016-08-20
DOC: add 3-clause BSD license.

rgommers:
Considered a CC license, but using the same license as for all other SciPy
code and documentation seems more appropriate.  Also less of an issue when
copying code/examples into an article.


ev-br:
Agreed, +1 for the same license everywhere.

rgommers:
no complaints, so just merging this

PR #3, opened 2016-10-15
DOC: update README with article types, repo organization, authorship policy

rgommers:


rgommers:
@stefanv I read the discussion on the scikit-image list about how you constructed the author list for your paper, but it was inconclusive. Looking back, any recommendations you would make?

stefanv:
@rgommers With that paper, we added everyone who contributed substantially to the paper as authors, and credited everyone else as "the scikit-image contributors".  The ordering of the list was more-or-less on a first come basis.

I guess some of us needed the paper more than others, hence the names, otherwise I may just have gone with "the scikit-image contributors".

rgommers:
Thanks Stefan. 

For the SciPy paper, I think we should go with everyone who contributed substantially to the project, not to the paper. The latter wouldn't be very practical even if we'd like that idea, given the large number of contributors.

ev-br:
@rgommers   do you want to add/change something further? I think we should merge this and iterate if necessary.

rgommers:
Let's do that!

Issue #8, opened 2018-02-05
set up paper build


rgommers:
We need:

- a Makefile to build the paper
- TravisCI or CircleCI to rebuild the paper on merges into or pushes to master.
- Uploading the built pdf to a fixed location, and link that from the README.

rgommers:
@person142 volunteered for this if we go with the LaTeX + Makefile option (thanks!), which it looks like we are.

rgommers:
Bonus points to implement the suggestion of @pv:

If you want to edit someone else's part, I think small changes should
be possible to do without asking. For larger changes, I suggest to add
instead a comment macro of the type `\PV{some comment}`, `\RG{some
comment}` in the text that explains the issue. Or, go with github issues
or PRs in those cases where you explicitly want feedback before
proceeding.

charris:
Thought I post this link about copyright because the issue is bound to come up at some point. https://academia.stackexchange.com/questions/7/does-publishing-a-paper-on-arxiv-prevent-me-from-submitting-it-to-a-non-open-acc

ilayn:
The macros are already implemented in [`changes`](https://tex.stackexchange.com/a/65466/3235) package.

ilayn:
Never had a problem of that sort but I don't have a strong opinion about the toolchain (yet). I am not sure if we would really need much of this kind comments anyhow if we keep the TeX files on GitHub and use the review system.  Especially if we separate things and make a master file with a lot of 

    ...
    \input{history}
    \input{modules}
    ...
etc. Then each section can be reviewed separately.

perimosocordiae:
For the Makefile, I'll recommend using `latexmk`: https://tex.stackexchange.com/a/40759/5105

I used it for several papers and my dissertation, with no complaints.

tylerjereddy:
+1 for `latexmk` as well; writing scripts that try to compile i.e., N times to get the citations right, etc., can get messy pretty quickly

person142:
Closing--done in gh-12.

PR #12, opened 2018-02-09
Makefile for SciPy 1.0 paper

person142:
Working on gh-8. Still need to hook things up to a CI, but it seems worthwhile to let people comment on the makefile first.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,7 @@
+**/*~
+
+**/*.aux
+**/*.fdb_latexmk
+**/*.fls
+**/*.log
+**/*.pdf
--- a/scipy-1.0/Makefile
+++ b/scipy-1.0/Makefile
@@ -1,8 +1,10 @@
 TEX = $(wildcard *.tex)
 
+.PHONY: paper
 paper: $(TEX)
 	latexmk -pdf -use-make paper.tex
 
+.PHONY: clean
 clean:
-	@rm *~
+	@rm -f *~
 	latexmk -CA
--- a/scipy-1.0/Makefile
+++ b/scipy-1.0/Makefile
@@ -1,0 +1,8 @@
+TEX = $(wildcard *.tex)
+
+paper: $(TEX)
+	latexmk -pdf -use-make paper.tex
+
+clean:
+	@rm *~
+	latexmk -CA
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -0,0 +1,7 @@
+**/*~
+
+**/*.aux
+**/*.fdb_latexmk
+**/*.fls
+**/*.log
+**/*.pdf
```
tylerjereddy: Good idea. There's a pretty well thought out `.gitignore` in the [official github gitignore repo for LaTex](https://github.com/github/gitignore/blob/master/TeX.gitignore) which I tend to pull from progressively as a work builds up in complexity / intermediate files generated.

tylerjereddy: We can probably safely add some ignore tags for popular editors too, but that burden doesn't all have to fall on this PR -- I'm sure one of us will eventually start seeing swap files and just patch the ignore file.

tylerjereddy: Have we decided on the citation approach yet? When we do that, there will be more stuff to add here as well.

rgommers:
Makefile works for me, thanks @person142

scipy-1.0/Makefile
```diff
@@ -0,0 +1,8 @@
+TEX = $(wildcard *.tex)
+
```
perimosocordiae: Worth adding: `.PHONY: paper clean`

.gitignore
tylerjereddy: Yeah, something like that.

person142:
Ok, I got CircleCI working in my fork and just pushed those changes here. I think an owner will have to turn on builds for this repository.

person142:
This was the result in my fork:

https://26-74767083-gh.circle-artifacts.com/0/papers/artifacts/paper.pdf

person142:
To keep things rolling, bump on this point:

> an owner will have to turn on builds for this repository.

rgommers:
@person142 done!

It said in the instructions:

    create a folder named .circleci and add a file config.yml (so that the filepath be in
    .circleci/config.yml).

but this should work as well

rgommers:
Okay that worked, can see the pdf under artifacts.

rgommers:
Okay works, so merging. Follow up on content of `.gitignore` etc can be tweaked later.

person142:
Thanks @rgommers.

PR #14, opened 2018-03-05
ENH: add document structure and Scientific Reports template.

rgommers:
This builds fine for me locally within my IDE (Kile), but not via the Makefile. I'm not a regular user anymore, so not sure what the best way to change the Makefile is. Ideally it should also handle having the cls/sty files in a subdirectory; e.g. move `paper.tex` and `references.bib` there before building. AFAIK in LaTeX it's not possible to include paths. Any suggestions?


rgommers:
I've got `naturemag-doi.bst`, think it's in `texlive-publishers` on Arch Linux. But happy to bundle it if others think that's better.

tylerjereddy:
Running `make` from the `scipy-1.0` folder on Ubuntu works just fine for me, except that I needed to manually add `naturemag-doi.bst`(and the original version I found online was causing issues -- seems there are better / worse versions of that floating around).

I see that moving the cls/sty files to a subdir causes make to break -- I'll see if I can get this to work locally.

tylerjereddy:
After moving the cls/sty files to a subdir, I could successfully run `make` by [modifying the build](https://tex.stackexchange.com/a/93733/43006) command to: `TEXINPUTS=.//: latexmk -pdf -use-make paper.tex`

This assumes that you will run make from the scipy-1.0 folder; it basically just uses that environment variable to recursively search the project folders for additional class/ style files after checking the conventional sources for this information in a typical latex install. I realize that you may want even more flexibility than that in calling make, but perhaps this is at least a start.

If you have a more specific scenario that needs to work I can try to look at it.

rgommers:
Thanks Tyler, that helped! I still need to run `make paper` three times to resolve references etc. - I'd have thought that `latexmk` would figure that out by itself. Does that work for you or do I just put the command in 3 times?

rgommers:
Okay CircleCI failed too, `xcolor.sty` missing. Sigh. The advice for CircleCI seems to be to build a Docker container with everything one needs ....

rgommers:
> perhaps we should make an official SciPy account on Docker hub.

That sounds potentially useful. For now just for this, but I can imagine it being used for dev setups etc.

I'm not sure what's the best approach to fix things here right now - add packages until things stop failing, or add a larger `texlive-extra` / `texlive-publishers` package. Things get cached on CircleCI after the first run, so I suspect the latter is less work and will be okay.

person142:
> or add a larger texlive-extra / texlive-publishers package. Things get cached on CircleCI after the first run, so I suspect the latter is less work and will be okay.

This is what I was thinking too. Leaving stuff out only reduced the container size by about 20%, so we weren't saving much anyway.

tylerjereddy:
> Thanks Tyler, that helped! I still need to run make paper three times to resolve references etc. - I'd have thought that latexmk would figure that out by itself. Does that work for you or do I just put the command in 3 times?

A single `make paper` works just fine for me, including references, even after a `git clean -xfd` to make sure I'm not leveraging any `.aux` files, etc. I've [opened a reference PR](https://github.com/rgommers/scipy-articles/pull/1) to your fork of scipy-articles just in case anything else might be slightly different on my end. I just run `make paper` in the `scipy-1.0` folder -- not sure if perhaps you are running it from somewhere else?

rgommers:
Fixed now, thanks Tyler!

rgommers:
That sounds like a good plan to me.

Issue #9, opened 2018-02-05
Scope and structure of SciPy 1.0 paper


rgommers:
**EDIT: tentative structure and authors:**

- [ ] Introduction *authors:*  @stefanv @jarrodmillman
  - [x] History 
  - [x] Project goals / scope of library 
  - [x] Current status (maturity, users)
  - [x] scikits
  - [x] timeline ([example](https://twitter.com/choldgraf/status/1009602557692375041?s=19))
- [x] Architecture and implementation choices *authors:* 
  - [x] Submodule organisation  *authors:*  @WarrenWeckesser 
  - [x] Common infrastructure (build, CI, etc.) **this may already be covered in test section below** 
  - [x] Language choices *authors:*  @tylerjereddy
  - [x] API and ABI evolution
- [x] Recent technical improvements:
  - [x] Data structures:
    - [x] `cKDTree` *authors*: @rainwoodman, @tylerjereddy 
    - [x]  `sparse` matrices *authors:* @perimosocordiae
  - [x] `LowLevelCallable`  *authors:* @tylerjereddy 
  - [x] `cython_blas/lapack/special` *authors:* @insertinterestingnamehere, @person142
  - [x] `optimize` (new optimizers, incl. benchmarks) *authors:* @antonior92, @mdhaber
  - [x] `stats` distributions *authors:* @josef-pkt
  - [x] `interpolate` (polynomial interpolators, `PPoly` & friends). *authors:* @ev-br
  - [x] test and benchmark suite *authors:* @tylerjereddy 
- [x] Project organisation and community *authors:*  @rgommers
  - [x] Governance (incl. CoC) *authors:*  @tylerjereddy
  - [x] Roadmap *authors:* @tylerjereddy 
  - [x] Community beyond the SciPy library (ecosystem, scipy.org, etc.) (perhaps covered in history?)
  - [x] Maintainers (i.e., core devs / commit rights) & contributors
    - [ ] may be useful: https://github.com/docathon/watchtower, https://github.com/machine-shop/oss_devkit/blob/master/plot_pr.py (PR over time plot, there may be other better versions out there)
- [ ] Conclusion  *authors:* @tylerjereddy
  - [x] impact now
  - [ ] future development (including key issues: sparse arrays @stefanv, ndimage pixel vs point @stefanv @jni ?, splines, fftpack vs. np.fft and linalg vs. np.linalg, under-maintained submodules)

**Original description**

We need to decide on the topics/structure/scope of this paper. 

[...]

PR #15, opened 2018-03-13
ENH: Early draft of asv benchmark suite subsection

tylerjereddy:
This PR contains a very early draft of the asv / benchmarking subsection, just to get things rolling there.

Various issues / thoughts:

- the current `asv` command used obviously is not getting enough successful benchmarks; failures happen in the older commits in the tested range; we might be able to make this look more reasonable by removing the requirement for i.e., merge commits only and adjusting the specific benchmark we probe (this starting command was chosen based on a quick parsing of various cKDTree improvements in the commit history)
- should probably add Python 3.6 to the test matrix as Python 2.x will indeed be quite close to EOL by the time the paper is out? (perhaps substitute in place of Python 2.x for clarity)
- there's probably various concerns related to where I've put the "raw image" file -- maybe should go somewhere else in repo
- we might want to include a sort of supplementary data section with the i.e., `asv.conf.json` that was actually used, or a table summarizing its parameter choices?
- would we prefer to actually commit the raw data for the benchmark numbers to increase the "executability" of the paper & allow customization of the plot appearance? (but do we really want to spend time writing plotting code for that JSON-formatted data on top of what `asv` can already do on its own?)


andyfaff:
Might be a good idea to submit all the code/notebooks to reproduce all the data/figures presented in the paper as supp info.

charris:
Could you clarify your suggestion regarding Python 3?

andyfaff:
> Could you clarify your suggestion regarding Python 3?

I think Tyler means at the moment the asv configuration uses 2.7. If a user wants to run 3.x, then that file has to be modified.

tylerjereddy:
> I think Tyler means at the moment the asv configuration uses 2.7. If a user wants to run 3.x, then that file has to be modified.

Yup

charris:
Don't see why we couldn't change that. @pv Thoughts?

andyfaff:
This is probably off topic, but there are a few tweaks that need to be made. Probably take out 2.7 and replace with 3.x, and remove the maintenance branch entries.
It would be useful if there was a computer somewhere which ran all benchmarks to keep track of them over time, on a single configuration. Not sure how much resources that would need.

tylerjereddy:
[Here's a link to the log (output) file from the benchmarks used here](https://github.com/scipy/scipy-articles/files/1811275/log.txt). More errors crop up in the older tests near the end of the file. This is fairly typical since call signatures and feature availability vary over time; it might be straighforward to adjust the benchmarks so that more of the history can be sampled, or it could be tricky. Depending on how tricky, this may affect the benchmark we ultimately decide to "feature" for demonstration purposes.

As @andyfaff notes, it may be possible to reduce the occurrence of benchmarks that lose their "backward compatibility" by having some semi-automated machine running them. Too time consuming for direct usage in CI though, but a number of projects do make some effort to do something of this nature.

perimosocordiae:
I'm pretty sure @pv has a machine that automatically runs benchmarks for every commit, because he runs https://pv.github.io/scipy-bench/

tylerjereddy:
Yeah, looks like a similar situation for cKDTree `time_query` there with fragility before ~ release 0.17.

andyfaff:
> I'm pretty sure @pv has a machine that automatically runs benchmarks for every commit, because he runs https://pv.github.io/scipy-bench/

That's one reason to keep the status quo, but it should probably be changed as others have to manually edit to run the benchmarks.

pv:
The machine that produces the pv.github.io results does not use that
asv.conf config file.

tylerjereddy:
The updated results with `boxsize` argument ignored and focusing on the L2 (Euclidean) norm only for cKDTree look perhaps a bit more promising with sampling back to a few more versions:

image

If people think this might be a promising candidate for `asv` usefulness exposition I could try repeating with these adjustments:

- increased sampling density
- use Python 3.6 instead of 2.7

tylerjereddy:
I note that @stsievert has also advocated for recent performance improvements in `convolve` and `correlate` (in case we're less impressed with what we're seeing with cKDTree so far, or want a more diverse exposition).

tylerjereddy:
For now, I'm repeating the cKDTree stuff with Python 3.6 and 5x increased sampling density; I'll aim to update with those results tomorrow to see if they are any more compelling.

tylerjereddy:
Here's the Python 3.6 results for time_query with higher sampling density, L2 norm only, and boxsize parameter ignored:

image

Looks a bit more compelling with the greater sampling of history now -- goes back to before 0.11. It would be nice to patch the benchmark failures in the gap region -- [the log file](https://github.com/scipy/scipy-articles/files/1834551/log2.txt) may have some useful clues.

stsievert:
I've done an initial speed test with the benchmarks from https://github.com/scipy/scipy-articles/issues/9#issuecomment-375849909 @rgommers I tested with asv yet.

It looks like the speed increase would be about 10x. I would need to make some small changes to the benchmarks as reflected in https://github.com/scipy/scipy/pull/8607. Here's the results I generated on my machine for testing `convolve` with `method=direct` (<v0.19) and `method=auto` (default for >=v0.19).

| Mode | `method=direct` | `method=auto` |
|:--:|:--:|:--:|
| full | 188.38ms | 14.21ms |
| valid | 5.44ms | 4.29ms|
| same | 67.07ms | 16.60ms |

person142:
p.s. finally updated the Docker container so the CI build is working again:

https://8-66140682-gh.circle-artifacts.com/0/papers/artifacts/paper.pdf

tylerjereddy:
If there's a general consensus that the benchmarks targeted by @stsievert would be preferred here I can repeat with those benchmarks once they are merged (it looks like we don't have a solid example that doesn't require at least some changes to be merged to the current suite), or I could close the PR if we'd prefer to start from scratch. Let me know.

rgommers:
thanks Josh

rgommers:
> If there's a general consensus that the benchmarks targeted by @stsievert would be preferred

not sure yet - they're well-known functions so if the graph of speed improvements looks better then it may be a good idea.

rgommers:
The `cKDTree` results look better now. I'd expect that with Python 2.7 you coud go further back and the cumulative improvements would look even better.

tylerjereddy:
Progress on the expanded benchmarks is around 39 % as of this morning--will update when they finish.

tylerjereddy:
Python 2.7 high density sampling of `cKDTree` `time_query` benchmark using Python 2.7 and the L2 (Euclidean) norm with boxsize ignored:

image

This goes back to ~version 0.7, which is better. For really old commits:
`global name 'cKDTree' is not defined`, which is fair enough

What I don't like is the gap in intermediate commits, where we get a Cythonization issue (see below the fold) -- is it worth investing the time to try and i.e., pin a different Cython version to remedy the gap?

<details>
```
[ 20.75%] · For scipy commit hash 6b36d00b:
[ 20.75%] ·· Building for conda-py2.7-Cython0.27.3-Tempita-blas-lapack-numpy1.8.2-pytest-six
[ 20.75%] ··· Error running /home/tyler/github_projects/scipy/benchmarks/env/3254d31fd71d64984dfa4052939c5dee/bin/python setup.py build
              STDOUT -------->
              Processing scipy/sparse/_csparsetools.pyx.in
              Processing scipy/sparse/csgraph/_tools.pyx
              Processing scipy/sparse/csgraph/_traversal.pyx
              Processing scipy/sparse/csgraph/_reordering.pyx
              Processing scipy/sparse/csgraph/_shortest_path.pyx
              Processing scipy/sparse/csgraph/_min_spanning_tree.pyx
              Processing scipy/ndimage/src/_ni_label.pyx
              Processing scipy/interpolate/interpnd.pyx
              Processing scipy/interpolate/_ppoly.pyx
              Processing scipy/io/matlab/mio_utils.pyx
              Processing scipy/io/matlab/streams.pyx
              Cythonizing sources 
              STDERR -------->
        
              Error compiling Cython file:
              ------------------------------------------------------------
              ...
        
                  cpdef int all_data_read(self):
                      return (self._max_length == self._read_bytes) and \
                             (self._buffer_size == self._buffer_position)
        
                  cpdef long int tell(self):
                       ^
              ------------------------------------------------------------
        
              streams.pyx:204:10: Signature not compatible with previous declaration
              Traceback (most recent call last):
                File "/home/tyler/github_projects/scipy/benchmarks/env/3254d31fd71d64984dfa4052939c5dee/project/tools/cythonize.py", line 198, in <module>
                  main()  
                File "/home/tyler/github_projects/scipy/benchmarks/env/3254d31fd71d64984dfa4052939c5dee/project/tools/cythonize.py", line 194, in main 
                  find_process_files(root_dir)
                File "/home/tyler/github_projects/scipy/benchmarks/env/3254d31fd71d64984dfa4052939c5dee/project/tools/cythonize.py", line 186, in find_process_files
                  process(cur_dir, fromfile, tofile, function, hash_db)
                File "/home/tyler/github_projects/scipy/benchmarks/env/3254d31fd71d64984dfa4052939c5dee/project/tools/cythonize.py", line 160, in process 
                  processor_function(fromfile, tofile) 
                File "/home/tyler/github_projects/scipy/benchmarks/env/3254d31fd71d64984dfa4052939c5dee/project/tools/cythonize.py", line 71, in process_pyx
                  raise Exception('Cython failed')
              Exception: Cython failed
              Traceback (most recent call last):
                File "setup.py", line 249, in <module>
                  setup_package()
                File "setup.py", line 242, in setup_package
                  generate_cython()
                File "setup.py", line 162, in generate_cython
                  raise RuntimeError("Running cythonize failed!")
              RuntimeError: Running cythonize failed! 
```

rgommers:
I think trying another Cython version will probably just give another issue at some other point in time. The best would be to not pin Cython to a single version for the whole range of commits, but use the version that we used for each release - that has the best chance of working. This probably needs a bit of digging to find those versions, plus some work to enable this when running the benchmarks. 

Would be useful in general, but not sure about the need for this paper.

tylerjereddy:
> best would be to not pin Cython to a single version for the whole range of commits, but use the version that we used for each release

Interesting, @pv has this been considered for incorporation to `asv` in the past? My guess is this might be tricky to implement & question of user-supplied "array" of Cython versions vs. auto-detection of the version "coupled" to a release.

Since that might be time consuming to develop, would the paper benefit from plotting the data with a sort of "axis break" / squiggly line style thing? I'm not sure that those kinds of axis breaks are broadly considered an improvement in clarity / professionalism though.

In the absence of actually incorporating the feature discussed above into asv, perhaps just digging up one or two more suitable Cython versions based on the previous releases and reapeating benchmarks with those might be suitable for "filling in the gap." This is a bit more brute force--re-running the full suite with an older Cython version so that a subset of commits can run (while the previously-successful commits may fail), but we can just combine that data as long as we are clear about what was done / same machine being used, etc.

stsievert:
I've made some of `Convolve` benchmarks more practical in https://github.com/scipy/scipy/pull/8607; now, they simulate convolving an image with a small kernel (e.g., 512x512 image with 5x5 kernel). This is common, both in simple and advance uses (e.g., blurring an image and edge detection).

screen shot 2018-03-30 at 2 03 28 pm

tylerjereddy:
Running with the older Cython version pinned has filled in most of the gaps (see below). I think this looks much better now -- maybe I'll just update the PR accordingly & then other compelling benchmarks might be added in as appropriate (or replace mine) in subsequent PRs. One more round of refinement with a third version of Cython might be called for if we eventually decide that this is one of the final benchmarks to make the cut--and perhaps a minor adjustment to make sure the x axis goes to 1.0 instead of stopping a little before.

image

tylerjereddy:
- updated PR to reflect the improved cKDTree asv benchmark results

- I had to pin the location of Table 1 from the optimize work to prevent its interjection into my subsubsection -- @antonior92 suggestions for reducing the impact of this on the formatting of the optimize material compiled here alongside asv stuff?

- I had the JSON config file in a supporting information path -- objections to the path I've chosen for this?

tylerjereddy:
Ok, the compiled PDF formatting looks much cleaner now with optimize and asv sections no longer interfering with each other / other sections. I'll aim to merge on Sunday if there are no objections before then; this doesn't mean that the asv section wont' still evolve or have a different final benchmark(s) in the end.

tylerjereddy:
Ok, in it goes to keep things moving along -- see #19 for additional discussion of the evolution of the benchmark section (this isn't final).

Issue #13, opened 2018-02-22
Recent technical improvements to scipy.optimize


mdhaber:
Since there have been many recent contributors to `scipy.optimize`, I think a distributed approach to drafting this section would work best. Once we collect components, those who are interested can work to edit this section to make it fit with the rest of the paper. Of course, contributors would understand that their components may need to be edited and might not even make the final paper due to space constraints, but that we'd really appreciate their input in any case.

I've quickly reviewed closed PRs labeled `scipy.optimize` from the past few years, looking mainly at those prefixed BUG, ENH, or WIP, and picked out a few that seem appropriate for the paper. Please forgive (and point out!) any omissions or incorrect attributions.

@nmayorov [#5556](https://github.com/scipy/scipy/pull/5556), [#5147](https://github.com/scipy/scipy/pull/5147)
@felixlen [#7292](https://github.com/scipy/scipy/pull/7292)
@larsmans [#5158](https://github.com/scipy/scipy/pull/5158)
@surhudm (or @nmayorov) [#6493](https://github.com/scipy/scipy/pull/6493)
@andyfaff [#4191](https://github.com/scipy/scipy/pull/4191), [#3446](https://github.com/scipy/scipy/pull/3446), [#5605](https://github.com/scipy/scipy/pull/5605)
@pv [#4392](https://github.com/scipy/scipy/pull/4392) (and potentially [#6769](https://github.com/scipy/scipy/pull/6769) / [#5536](https://github.com/scipy/scipy/pull/5536) if you think they're significant)
@person142 [#5557](https://github.com/scipy/scipy/pull/5557)
and of course
@antonior92  [#6919](https://github.com/scipy/scipy/pull/6919), [#7165](https://github.com/scipy/scipy/pull/7165)
myself, [#7123](https://github.com/scipy/scipy/pull/7123)

Here is a draft about the new `linprog` interior-point method. We might decide we want more or less about different topics - or even that we want something pretty different from what I have below - but I thought it would be helpful to give _something_ as an example.

_A new interior-point optimizer for continuous linear programming problems, `linprog` with `method='interior-point'`, was released with SciPy 1.0. Implementing the core algorithm of the commercial solver MOSEK [1], it solves all of the 90+ Netlib LP benchmark problems tested [2]. (Add something about speed benchmarks?) Unlike some interior point methods, this homogenous self-dual formulation provides certificates of infeasibility or unboundedness as appropriate. (Add something about infeasible/unboundedness benchmarks?)

[...]

PR #16, opened 2018-03-26
[WIP] ENH: Optimize Section

mdhaber:
Still don't know what I'm doing with git. Just trying to figure out how to add my contributions to the paper.

I'd appreciate feedback to any questions asked in the text.

### DIFF
--- a/scipy-1.0/paper.tex
+++ b/scipy-1.0/paper.tex
@@ -57,6 +57,7 @@ LowLevelCallable
 \subsection*{Cython bindings for BLAS, LAPACK and special}
 
 \subsection*{Numerical optimization}
+\input{scipy-optimize}
 
 \subsection*{Statistical distributions}
 
--- a/scipy-1.0/references.bib
+++ b/scipy-1.0/references.bib
@@ -20,3 +20,50 @@
   publisher={Inderscience Publishers}
 }
 
+@article{kuhn1955hungarian,
+  title={The Hungarian method for the assignment problem},
+  author={Kuhn, Harold W},
+  journal={Naval Research Logistics (NRL)},
+  volume={2},
+  number={1-2},
+  pages={83--97},
+  year={1955},
+  publisher={Wiley Online Library}
+}
+
+@article{kuhn1956variants,
+  title={Variants of the Hungarian method for assignment problems},
+  author={Kuhn, Harold W},
+  journal={Naval Research Logistics (NRL)},
+  volume={3},
+  number={4},
+  pages={253--258},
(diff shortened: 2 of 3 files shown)


mdhaber:
The most important questions right now:

- Is this an appropriate way of breaking down the optimization subsection?
- Does the introductory text written for root finding provide a good template for the introductionar material of the remaining subsubsections? That is, should I introduce define all the problems mathematically before they are discussed in more detail, or are we assuming that the reader is familiar with these standard types of problems?

antonior92:
> Is this an appropriate way of breaking down the optimization subsection?

I liked it. It gives a good notion of what needs to be done. Maybe we can merge a few subsections later on.

> Does the introductory text written for root finding provide a good template for the introductionar material of the remaining subsubsections? That is, should I introduce define all the problems mathematically before they are discussed in more detail, or are we assuming that the reader is familiar with these standard types of problems?


I think we should try to define mathematically the most important problems.

Since the paper will discuss very diverse topics, it is very likely that the reader does not know deeply about many of them. So I think we should not expect that the reader has a background on the subject. Maybe we could try to delegate more complicated concepts (like *local* and *global convergence* in my PR) to tables, figures and footnotes.

rgommers:
In general I think that there are too many sub-sections here. We shouldn't try to cover all the functionality that `optimize` has to offer, just focus on the highlights / recent improvements (which I'd say are the methods behind `minimize` and `linprog` only).

mdhaber:
@antonior92 I want to leave out the description of the problem for now. The reason is that I want to put the content of #17 alongside this so we can better judge what sorts of things we like about each and decide how to unify them, as the fact that #17 has the mathematical problem statement and #16 doesn't is just one of many significant differences. If that sounds good, please merge this and #17. Then, I'll write a post in #13 listing these differences and asking for feedback on how we can resolve them so that we have a more unified paper.

mdhaber:
@rgommers in #13 I had a table of "recent" improvements that I thought we might include. Based on that, @surhudm had contributed a bit about `curve_fit`, @andyfaff was going to write about global optimizers, I've been trying to contact @larsmans about linear sum assignment, and I was considering asking @pv about the addition of `df-sane` to `root`. Should these be left out?

rgommers:
> ...  Should these be left out?

I think so. Just based on space requirements: if we include all those types of things, we're going to end up with a 30+ page paper (which won't be acceptable to the journal very likely). They would be great for a more in-depth article on `scipy.optimize`, but here I think we should focus on a selection of the most important changes only.

PR #17, opened 2018-03-27
Add subsection about `minimize` function

antonior92:
Hi all,

I just wrote a first draft of the table+text for ``minimize`` function. I wrote a full paragraph about ``trust-exact`` and another about ``trust-krylov`` (which are the methods added more recently). And I tried to give a general overview of all methods in comparative terms.

@felixlen can you help me fiding out the local convergence rate of ``trust-krylov`` and if there are guarantees of **global convergence** for it? I was not able to find this information by myself for this method. For the other unconstrained minimisation methods Nocedal book or original papers usually contained this kind of information.

@rgommers I added a table entry about the version each optimization method was added to scipy. I was not able to find out when some of the early methods were added, so I just wrote ``<0.7``. Where can I look for this information?

If we have space, I think it would be nice to include some benchmark results in this section latter on.


scipy-1.0/minimize.tex
```diff
@@ -0,0 +1,118 @@
...
+  \end{tabular}
+  \caption{Optimization methods from \texttt{minimize} for unconstrained minimization.  The field \textit{It is a wrapper?}
+    informs if the implementation available in SciPy is a wrapper to a function written in other programming language
+    (e.g. C or FORTRAN). The fields \textit{\nth{1}} and \textit{\nth{2} derivatives}
+    informs if the first or second order derivatives are required. When \textit{\nth{2} derivatives} is flagged
+    with the word \textit{Approx.} it means that, while the algorithm does not requires the second-order derivatives from
+    the user, it computes an approximation internally and uses it to accelerate the method convergence.
+    \textit{Local convergence} gives
+    a lower bound to the rate of convergence of the iterations sequence once the iterate is close enough to the
+    objective function solution. The convergence rates mentioned at the table
```
mdhaber: "objective function solution" -> "solution", right? Literally, a "function" doesn't have a "solution" unless by solution you mean "root". You mean the solution to the optimization/nonlinear programming problem, and I think that "solution" is clear enough.
"mentioned at the table" doesn't sound natural (again, to a A-EN speaker). It would be "mentioned in the table", but "mentioned" is probably not the right word for technical writing. I'd suggest just dropping this phrase entirely: "The convergence rates are, from slowest to fastest:"

scipy-1.0/minimize.tex
```diff
@@ -0,0 +1,118 @@
...
+    are, from the slowest to the fastest:  linear (L), superlinear (S) and quadratic (Q). The \textit{global convergence}
+    is marked for the algorithms for which there are some proof it always converge to some stationary
```
mdhaber: "proof it always converge to" -> "proof of convergence to"

scipy-1.0/minimize.tex
```diff
@@ -0,0 +1,118 @@
...
+    point (i.e. a point $x^*$ for which $\nabla f(x^*) = 0$).}
+  \label{tab:minimize-unconstrained}
+\end{table}
+
+
+The \texttt{minimize} function provides methods for solving general non-linear optimization
```
mdhaber: "non-linear" -> "nonlinear" because "nonlinear" is a real word.

scipy-1.0/minimize.tex
```diff
@@ -0,0 +1,118 @@
...
+problems. It unifies several methods under the same general interface in order to facilitate
```
mdhaber: consider "problems." -> "problems commonly referred to as nonlinear programming problems."
consider "under the same general interface" -> "with a common interface"

scipy-1.0/minimize.tex
```diff
@@ -0,0 +1,118 @@
...
+interchanging between solvers with different characteristics.  Table~\ref{tab:minimize-unconstrained} summarizes
```
mdhaber: I think technically "between" is used when there are _two_ things and "among" is for three or more. Then again, I guess a transition is always from one to a second, so maybe this a fine.
However, I would also reword this to give more information - answering the question that one might ask: "why would I need to switch methods?"
Maybe "common interface to facilitate solution of the problem using different methods, as finding the most effective algorithm for solving a given problem is often a matter of trial and error."

scipy-1.0/minimize.tex
```diff
@@ -0,0 +1,118 @@
...
+all the methods available for solving unconstrained minimization problems of the type:
+\begin{equation}
+  \label{eq:unconstrained-minimization-prob}
+  \min_x f(x)
+\end{equation}
+where $f$ is a multivariate function $f: \mathbb{R}^n \rightarrow \mathbb{R}$ and $x \in \mathbb{R}^n$.
```
mdhaber: I suggest we add the command \RR for \mathbb{R} as this will be used all over.
Also, to indicate multivariate/vector valued entities, do we want to use boldface? I did, but I can take it out. In any case, we should decide now.

scipy-1.0/minimize.tex
```diff
@@ -0,0 +1,118 @@
...
+
+There are a sufficient variety of methods to enable the user to solve problems with the diverse properties.
```
mdhaber: "are" -> "is" because "a ... variety" is actually singular
For instance, it would be "There are several methods...", but it is "There is a variety..."
"with the diverse properties" is vague. I know what you mean, and I'm not sure how to say it better... something to think about. In any case, remove "the"; you don't need the "definite article" here.

scipy-1.0/minimize.tex
```diff
@@ -0,0 +1,118 @@
...
+Derivative-free methods (e.g. \texttt{Nelder-Mead} e \texttt{Powell}) are appropriate for minimizing noisy objective
+functions. When the derivatives are available, on the other hand, methods with faster convergence can also be used.
```
mdhaber: "the derivatives" -> "clean derivatives"?  Here I'm using "clean" to indicate that I mean in the case when the function is not noisy. Or actually, it would probably be better if you wrote "are appropriate for minimizing non-differentiable or noisy functions objective functions. When derivatives are available..."
" methods with faster convergence can also be used." -> "other methods offer faster convergence."

larsoner:
> I was not able to find out when some of the early methods were added, so I just wrote <0.7

Apparently 0.7 came out in 2009, so I think it's okay to omit this information. If you are strongly motivated, 0.7 and 0.6 at least have source archives you could peruse here:

https://sourceforge.net/projects/scipy/files/scipy/

scipy-1.0/minimize.tex
```diff
@@ -0,0 +1,118 @@
...
+
+There are algorithms implementing two competing strategies: line-search and trust-region. The so-called line-search (LS)
+methods choose a direction and then find the step length that gives the (approximately) optimal point in the given
+direction. Trust-region (TR) methods, on the other hand, fixes the maximum step length for which the local approximation
+is valid and finds approximately the best point inside this \textit{trust-region}.
+
```
mdhaber: These descriptions are great if the reader understands that all the algorithms are iterative, and this means that given a current "guess" you compute a new "guess" by adding to it a new vector, defined by direction and magnitude (step length). However, if the reader doesn't know that, then this description is not self-contained. I would suggest that if the reader doesn't know what "direction" and "step length" are, they won't understand the explanation, and that if they do know what "direction" and "step length" are, they don't need the explanation. We need to decide who the audience is or make this self-contained.

In my description of the "interior-point" method, I assumed a _lot_ of knowledge in the interest of compactness.  Once we decide on this, we'll need to be consistent.

scipy-1.0/minimize.tex
```diff
@@ -0,0 +1,118 @@
...
+An example of trust-region method introduced in the latest releases is \texttt{trust-exact}, which a  provides a fast convergence rate by solving 
+ almost exactly the trust-region subproblem. Its downside is it also require the Hessian
```
mdhaber: "require" -> "requires"

scipy-1.0/minimize.tex
```diff
@@ -0,0 +1,118 @@
...
+to be stored and factored every iteration, which may renders impossible the solution of
```
mdhaber: "may renders impossible" would be "may render impossible", but I suggest a different wording: "which may preclude the solution..."

scipy-1.0/minimize.tex
```diff
@@ -0,0 +1,118 @@
...
+very large problems.
+
+Some available unconstrained methods are appropriate for large-scale problems
```
mdhaber: Is a method of unconstrained problems commonly called an "unconstrained method"?

scipy-1.0/minimize.tex
```diff
@@ -0,0 +1,118 @@
...
+(i.e. problems with more than 10000 variables).
```
mdhaber: "i.e." means literally "that is"
This would suggest that there is a strict definition of "large-scale" problems - those with 10,000 variables or more.
Of the common Latin abbreviations, I would use "e.g." (for example) here.

scipy-1.0/minimize.tex
```diff
@@ -0,0 +1,118 @@
...
+\texttt{CG} method does not use second derivatives and is adequate to large-scale problems due to the low storage requirements.
```
mdhaber: Nowhere else is the method name followed by "method".
For the sake of consistency, "\texttt{CG} method does"-> "\texttt{CG} does"
"adequate to large-scale" -> "adequate for large-scale"

scipy-1.0/minimize.tex
```diff
@@ -0,0 +1,118 @@
...
+\texttt{Newton-CG}, \texttt{trust-ncg} and \texttt{trust-krylov} use the second order derivatives in order to get faster
```
mdhaber: "use the second order derivatives in order to get faster"->"use second order derivatives to achieve faster"

scipy-1.0/minimize.tex
```diff
@@ -0,0 +1,118 @@
...
+convergence but are still appropriate for large-scale problems because they factorate the Hessian in a inexact and iterative way,
```
mdhaber: "factorate"->"factorize"

scipy-1.0/minimize.tex
```diff
@@ -0,0 +1,118 @@
...
+which does not require the Hessian to be actually stored.
```
mdhaber: "actually stored" -> "stored explicitly"

scipy-1.0/minimize.tex
```diff
@@ -0,0 +1,118 @@
...
+
+The method \texttt{trust-krylov}, introduced in the latest release, presents itself as a good compromise between
```
mdhaber: "presents itself as" -> "is"

scipy-1.0/minimize.tex
```diff
@@ -0,0 +1,118 @@
...
+\texttt{trust-ncg} and \texttt{trust-exact}: it provides an more accurate solution to the trust-region
+subproblem than \texttt{trust-ncg} but does not require the Hessian to be stored  and factored as in \texttt{trust-exact}.
+
+Besides unconstrained problems, the \texttt{minimize} function also provides methods for solving
+minimization problems subject to constraints. Table~\ref{tab:minimize-constrained} summarizes all
+constrained optimization methods available, with information of which type of
```
mdhaber: "with information of which type of constraints" -> "including which constraint types"

scipy-1.0/minimize.tex
```diff
@@ -0,0 +1,118 @@
...
+constraints they are able to deal with.
+
```
mdhaber: Do you intend to add more about these? This is a lot shorter than the part about the techniques for unconstrained problems.

scipy-1.0/minimize.tex
```diff
@@ -0,0 +1,118 @@
...
+    \hline
+    Equality constraint: $c_{\text{eq}}(x) = 0$ &  &  &  & \cmark \\
+    \hline
+    Inequality constraint: $c_{\text{ineq}}(x) \ge 0$ &  &  & \cmark & \cmark \\
+    \hline
+    References & \cite{byrd_limited_1995, zhu_algorithm_1997}  & \cite{nash_newton-type_1984} &
+     \cite{powell_direct_1994, powell_direct_1998, powell_view_2007} & \cite{kraft_software_1988} \\
+    \hline
+  \end{tabular}
+  \caption{Optimization methods from \texttt{minimize} for constrained minimization.}
```
mdhaber: I would add a note about "COBYLA", that bound constraints $x^l \le x \le x^u$ can be expressed as inequality constraints $x \ge x^l$,  $-x \ge -x^u$ and equality constraints $c_{\text{eq}}(x) = 0$ can be expressed as $c_{\text{eq}}(x) \ge 0$ and $-c_{\text{eq}}(x) \ge 0$.
I don't understand why this isn't done automatically by `minimize` personally. Is it horrendously bad for the problem structure - so bad that it's not worth supporting (with a warning in the documentation)?

scipy-1.0/minimize.tex
felixlen: I think "lowest to fastest" is more appropriate than "slowest to fastet" und usually used when talking about convergence rates. The ultimate speed also depends on the constant and the cost per iteration.

scipy-1.0/minimize.tex
felixlen: "e" --> "and". Also you might want to mention that derivative-free methods usually are heuristics without any guarantee on solution quality.

scipy-1.0/minimize.tex
felixlen: If direct linear algebra is used (as is in your exact method), you'll want to switch to large scale probably earlier, say 1000 variables...

scipy-1.0/minimize.tex
felixlen: "a exact" -> "an exact"

scipy-1.0/minimize.tex
felixlen: "an more" -> "a more"

scipy-1.0/minimize.tex
```diff
@@ -0,0 +1,118 @@
+\subsubsection*{Solving nonlinear optimization problems}
+
+\begin{table}[]
+  \centering
+  \begin{tabular}{cccccccccc}
+      & \rotatebox{60}{\texttt{Nelder-Mead}} & \rotatebox{60}{\texttt{Powell}} & \rotatebox{60}{\texttt{CG}} & \rotatebox{60}{\texttt{BFGS}}& \rotatebox{60}{\texttt{Newton-CG}} & \rotatebox{60}{\texttt{dogleg}} & \rotatebox{60}{\texttt{trust-ncg}} & \rotatebox{60}{\texttt{trust-exact}} & \rotatebox{60}{\texttt{trust-krylov}} \\
```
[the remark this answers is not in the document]
antonior92: Yes! Thank you for finding the typos.

scipy-1.0/minimize.tex
```diff
@@ -0,0 +1,118 @@
...
+    \hline
+    Version added & $<$0.7 & $<$0.7 & $<$0.7 & $<$0.7 & $<$0.7 & 0.13 & 0.13 & 0.19 & 1.0 \\
+    \hline
+    It is a wrapper? &  &  & & &  &  & & & \cmark \\
```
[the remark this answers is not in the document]
antonior92: Ok I will change it to *Wrapper*.

scipy-1.0/minimize.tex
```diff
@@ -0,0 +1,118 @@
...
+    \hline
+    \nth{1} derivatives &  & & \cmark & \cmark & \cmark & \cmark & \cmark & \cmark & \cmark \\
+    \hline
+    \nth{2} derivatives &  &  &  & Approx. & \cmark & \cmark & \cmark & \cmark & \cmark \\
```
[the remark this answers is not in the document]
antonior92: Ok! I will use ~

scipy-1.0/minimize.tex
```diff
@@ -0,0 +1,118 @@
...
+    \makecell{Line-search (LS) or\\ trust-region (TR)} & Neither  & LS & LS & LS & LS & TR & TR & TR & TR \\
+    \hline
+    References & \cite{nelder_simplex_1965, wright_direct_1996} & \cite{powell_efficient_1964} &
+      \cite{polak_note_1969, nocedal_numerical_2006} & \cite{nocedal_numerical_2006} & \cite{nocedal_numerical_2006}  &
+      \cite{powell_new_1970, nocedal_numerical_2006} &  \cite{steihaug_conjugate_1983, nocedal_numerical_2006} &
+      \cite{conn_trust_2000, more_computing_1983} & \cite{gould_solving_1999, lenders_trlib:_2016} \\
+    \hline
+  \end{tabular}
+  \caption{Optimization methods from \texttt{minimize} for unconstrained minimization.  The field \textit{It is a wrapper?}
+    informs if the implementation available in SciPy is a wrapper to a function written in other programming language
```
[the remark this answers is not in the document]
antonior92: Ok!

scipy-1.0/minimize.tex
```diff
@@ -0,0 +1,118 @@
...
+    (e.g. C or FORTRAN). The fields \textit{\nth{1}} and \textit{\nth{2} derivatives}
+    informs if the first or second order derivatives are required. When \textit{\nth{2} derivatives} is flagged
+    with the word \textit{Approx.} it means that, while the algorithm does not requires the second-order derivatives from
+    the user, it computes an approximation internally and uses it to accelerate the method convergence.
+    \textit{Local convergence} gives
+    a lower bound to the rate of convergence of the iterations sequence once the iterate is close enough to the
```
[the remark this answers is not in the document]
antonior92: ok!

scipy-1.0/minimize.tex
antonior92: Ok

scipy-1.0/minimize.tex
antonior92: Ok

scipy-1.0/minimize.tex
antonior92: I borrow the notation from Nocedal book, which is the reference used for most of the methods.
He opts to not use boldface for vectors. I think 

I think I prefer this notation as it is, it helps keeping things simple. But I think we should wait and see how the remaining sections will choose to deal with vectors.

scipy-1.0/minimize.tex
antonior92: Ok thanks!

I agree with you that "diverse properties" is a little vague. Nevertheless, since I describe some properties latter on, I think it is ok...

scipy-1.0/minimize.tex
antonior92: I reformulated the phrase as: "Derivative-free methods (e.g. \texttt{Nelder-Mead} and \texttt{Powell}) are appropriate for minimizing non-differentiable or
noisy objective functions. When the derivatives are available, however, other methods offer faster convergence and have guarantees of convergence."

I think the fact that derivative-free methods do not have guarantees of convergence is implicit from the fact "other methods ... have guarantees of convergence"

antonior92: > "e" --> "and"

Sorry about that. "e" is the word for "and" in portuguese. I mixed things up.

scipy-1.0/minimize.tex
antonior92: Ok, I removed the word "step-length".  I still assume the reader understand what "direction" means (at least in some intuitive sense).

scipy-1.0/minimize.tex
antonior92: ok

scipy-1.0/minimize.tex
antonior92: Ok! Thanks

scipy-1.0/minimize.tex
antonior92: I think not hahaha

I will use "available unconstrained optimization methods".

scipy-1.0/minimize.tex
antonior92: >  I would use "e.g." (for example) here.

Ok. I agree with you that we do not want to sound like inventing a definition to large scale problems.

> If direct linear algebra is used (as is in your exact method), you'll want to switch to large scale probably earlier, say 1000 variables...

Ok.

scipy-1.0/minimize.tex
antonior92: ok

scipy-1.0/minimize.tex
antonior92: ok

scipy-1.0/minimize.tex
antonior92: ok

scipy-1.0/minimize.tex
antonior92: I choose to just remove the word "actually". I think it sounds better: "...which does not require the Hessian to be stored."

scipy-1.0/minimize.tex
antonior92: ok

scipy-1.0/minimize.tex
antonior92: ok

scipy-1.0/references.bib
```diff
@@ -20,3 +20,244 @@ @article{peterson2009f2py
...
+  pages = {1190--1208}
+}
+
+@article{lenders_trlib:_2016,
+  title = {Trlib: {{A}} Vector-Free Implementation of the {{GLTR}} Method for Iterative Solution of the Trust Region Problem},
+  timestamp = {2018-03-26T20:21:16Z},
+  journal = {arXiv preprint arXiv:1611.04718},
+  author = {Lenders, Felix and Kirches, Christian and Potschka, Andreas},
+  year = {2016}
+}
```
felixlen: The paper is now published, it would be great if you could update the reference:

```
  title         = {trlib: {A} vector-free implementation of the {GLTR} method for iterative solution of the trust region problem},
  author        = {F. Lenders and C. Kirches and A. Potschka},
  journal       = {Optimization Methods and Software},
  publisher     = {Taylor & Francis},
  year          = {2018},
  doi           = {10.1080/10556788.2018.1449842},
  url           = {https://doi.org/10.1080/10556788.2018.1449842}
```

scipy-1.0/minimize.tex
mdhaber: This is already much better. Conisder "other methods offer faster convergence and have guarantees of convergence" -> "other methods offer faster and guaranteed convergence."

scipy-1.0/minimize.tex
antonior92: Ok, I decided to merge both tables into one and talk about constrained and unconstrained algorithms together.

scipy-1.0/minimize.tex
antonior92: Ok, I decided to merge both tables into one and talk about constrained and unconstrained algorithms together.

It makes more sense.

scipy-1.0/minimize.tex
antonior92: Ok. I decide to proceed as follows:
1. I will say that COBYLA can deal with bound constraints. The user just have to do this trick you just said.
2. I still say it cannot deal with equality constraints. Since, expressing equality constraints as $c_{\text{eq}}(x) \ge 0$ and $-c_{\text{eq}}(x) \ge 0$ usually does not work well in practice.

scipy-1.0/references.bib
antonior92: Thank you. I will update the reference.

antonior92:
Thank you a lot @mdhaber  and @felixlen. I tried to address all your comments.

A few comments: 
1. I decided to merge the two tables into a single one.
2. I included the convergence rate for ``trust-krylov``. About @felixlen statement: ``It can be made quadratic by mandating less inexact solutions, but this typically requires more inner iterations and more hessian vector products.``. This is also true for ``Newton-CG``, ``trust-ncg`` and ``TNC``. I added a note about that as well.
3. Local and global convergence properties of ``SLSQP`` are not described anywhere I could find. So I left them empty. By my recent experience ``SLSQP`` can fail in very simple bidimensional problems and does not seems to be very robust.

felixlen:
@antonior92 Thanks for your work, will have a closer look later. Regarding 3. `SLSQP`: I scanned the reference describing `SLSQP` some time ago and made it available as gist [https://gist.github.com/felixlen/784792d231ce35a4ade2955dc71ded06](Kraft1988). `SLSQP` implements a method proposed by Schittkowski ("The nonlinear programming method of Wilson, Han, and Powell with an augmented Lagrangian type line search function", [https://link.springer.com/article/10.1007/BF01395810](Part 1), [https://link.springer.com/article/10.1007/BF01395811](Part 2)) with convergence analysis in [https://www.tandfonline.com/doi/abs/10.1080/02331938308842847]("On the convergence of a sequential quadratic programming method with an augmented lagrangian line search function").

The convergence analysis yields global convergence under standard assumptions and local superlinear convergence. I support your observations on the practical behaviour of `SLSQP`, having seen multiple instances where it failed or claimed convergence without being in a solution. I think the lack of Lagrange multipliers to check KKT a posteriori is a real problem...

mdhaber:
@antonior92 great idea to merge the tables, if it all fits.

rgommers:
That looks great Antonio!

For those who are looking for the rendered table: https://13-66140682-gh.circle-artifacts.com/0/papers/artifacts/paper.pdf

antonior92:
Ohh, I forgot to say:

@felixlen while I manage to get all the references, the link to the gist you have indicated is broken. Have you made it public?

felixlen:
@antonior92 Sorry! It is public, but there was a typo in the link. I have updated the comment, please try again.

antonior92:
Thanks Felix! I managed to access it now.

mdhaber:
@antonior92  Currently I think more important than fine tuning either this PR or #16 is resolving the different approaches between the two. I suggest we merge this and #16 so we can work on them side by side. If that sounds good to you, would you go ahead and merge both?

antonior92:
Ok, just merged #16. I will adatp this PR accordingly and merge it as well

PR #20, opened 2018-04-09
MAINT: global optimizer

andyfaff:
@mdhaber this may be of use. It can be expanded if necessary.


scipy-1.0/scipy-optimize.tex
```diff
@@ -130,3 +130,7 @@ \subsubsection*{Nonlinear Optimization}
   \end{tabular}
   \label{tab:minimize}
 \end{table}
+
+% Some problems may require the use of global optimizers to find the minimum of the objective function over the set of all parameter values, especially if the objective function possesses many local minima that other solvers could get stuck in.
+% \texttt{Basinhopping} \cite{Wales1997} uses random perturbation of the parameter vector, with local minimization. The acceptance of the new coordinates is dictated by the Metropolis test typically used in Monte Carlo simulations.
+% \texttt{differential\textunderscore evolution} uses
```
tylerjereddy: is this an incomplete comment / thought about `differential_evolution`?

andyfaff: It's not an incomplete thought. I wasn't sure if there was going to be space in the article, so I left it commented out. I'm happy if it is added to the manuscript.

andyfaff: It looks like it's missing half the text on DE!

andyfaff: In fact it's missing half the text I wrote. Not sure what went on with my text editor. I'll have to rewrite the bits.

mdhaber:
@andyfaff Thanks! I will resolve the conflicts I just introduced by accidentally pushing to the main repo master.

scipy-1.0/scipy-optimize.tex
```diff
@@ -130,3 +130,10 @@ \subsubsection*{Nonlinear Optimization}
   \end{tabular}
   \label{tab:minimize}
 \end{table}
+
+Some problems may require the use of global optimizers to find the minimum of the objective function over the set of all parameter values, especially if the objective function possesses many local minima that other solvers could get stuck in.
```
antonior92: I think it is worth mentioned that this global solvers are made available apart from the ``minimize`` interface.
Since we were talking about the minimize interface before, this may generate some confusion, if not explicitly mentioned.

antonior92:
Another point that may generate confusion is the use of ``global convergence`` in the table. 

It is worth including a note explaining global convergence do **not** mean convergence to the global minimum.

antonior92:
Finally, I think it is worth mentioning that global optimization algorithms are usually appropriate for problems with less than 100 variables, while the local optimization problems can deal with larger problems (to explain the purpose of local optimization algorithms).

andyfaff:
There were some merge issues, which is going to be a recurring problem with something like a paper. It's going to be better to merge quickly and merge often, otherwise too much bandwidth is going to be wasted. We can alway do the editing afterwards.

andyfaff:
I'm going to merge the PR now, I've got no issues with @antonior92 suggestions, they can be made in a further edit.

Issue #19, opened 2018-04-07
Decide on benchmark to show for ASV subsection


stsievert:
@tylerjereddy has drafted the ASV benchmark section in #15. In https://github.com/scipy/scipy-articles/pull/15#issuecomment-379328224, @tylerjereddy mentioned

>  I'll aim to merge on Sunday ... this doesn't mean that the asv section wont' still evolve or have a different final benchmark(s) in the end.

I think it's fine to merge that PR, but I would like to discuss which benchmark to display. Both @tylerjereddy and I have put in time to benchmark cKDTree and convolve respectively.

stsievert:
I think we should consider what we want the benchmark to show. I think the main goal is to show ASV's capabilities, but I also think we should show

* that benchmarks improve with increasing versions
* that the benchmarks are practical

What else do we want to show?

tylerjereddy:
@stsievert I went ahead with the merge since you were ok with it & there were no additional objections. Hopefully the section will continue to improve & what I've done could always be shifted to the supporting info over time.

stsievert:
Thanks for merging it. Do you have any comments on the goals mentioned above?

tylerjereddy:
Those are sensible goals -- I guess we can work on making our plots look as good as possible -- patching in those holes by running with 1 or 2 more versions of Cython for example.

We could also think about combining two different benchmarks in a single plot, though this may require i.e, custom code to work on the JSON if we want it to work nicely (and we may also simply prefer not to do a custom plot--since the default plots showcase the built-in plotting / comparison power of asv).

One thing I don't like about my contributed benchmarks is that you actually have to modify the source of the asv benchmark itself (not just the config file) to sample that far back in history.

I wonder if showcasing an example of `asv continuous` might also have some value -- to show that we can evaluate a single PR vs. the commit hash of the head of master -- this has occasionally been used in the past on isolated PRs to evaluate performance (though we often seem to accept i.e., informal benchmarks in a Jupyter nb as well...).

stsievert:
> to show that we can evaluate a single PR vs. the commit hash of the head of master

That'd be a good idea show highlight a particular improvement, though the value added isn't clear to me. I would like to show some of ASV's web interface: I've found it useful already.

I think we could express the benefits of `asv continuous` in the figure caption or text, especially if a very sudden improvement shown ("The improvement shown in Figure X could also be shown with `asv continuous` which would produce a bar chart comparison between two commits or branches").

[...]

PR #23, opened 2018-04-12
MAINT: add paragraph on `asv continuous`

stsievert:
Suggested in https://github.com/scipy/scipy-articles/issues/19#issuecomment-380520085

### DIFF
--- a/scipy-1.0/paper.tex
+++ b/scipy-1.0/paper.tex
@@ -88,15 +88,17 @@ LowLevelCallable
     confirming that performance generally remains constant or improves over the commit hash history of the project allows us to objectively
     measure that our code base is improving, to empower scientific applications.
 
-    Consider the following asv benchmark results, spanning roughly nine years of project history, that demonstrate the gradual performance
-    improvements in \texttt{scipy.spatial.cKDTree.query()} (nearest-neighbor search) performance, using command:
+    Consider the following asv benchmark results, spanning roughly nine years of project history. These demonstrate the gradual performance
+    improvements in a nearest-neighbor search, \texttt{scipy.spatial.cKDTree.query()} using the command:
 
     \texttt{python run.py run -e -s 800 --bench "\textbackslash btime\_query\textbackslash b" "02de46a546..b3ddb2c"}
+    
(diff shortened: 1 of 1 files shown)


scipy-1.0/paper.tex
```diff
@@ -88,15 +88,17 @@ \subsection*{Test and benchmark suite}
...
     measure that our code base is improving, to empower scientific applications.
 
-    Consider the following asv benchmark results, spanning roughly nine years of project history, that demonstrate the gradual performance
-    improvements in \texttt{scipy.spatial.cKDTree.query()} (nearest-neighbor search) performance, using command:
+    Consider the following asv benchmark results, spanning roughly nine years of project history. These demonstrate the gradual performance
+    improvements in a nearest-neighbor search, \texttt{scipy.spatial.cKDTree.query()} using the command:
 
     \texttt{python run.py run -e -s 800 --bench "\textbackslash btime\_query\textbackslash b" "02de46a546..b3ddb2c"}
+    
+   Any pull request can be compared against the \texttt{master} branch with the command \texttt{asv continuous master new-feature}. This will provide a benchmark against the master branch and the branch a new feature is implemented in. More features are available in the documentation, including arguments to select which benchmarks to run.
```
tylerjereddy: Perhaps move this paragraph so that it doesn't break up the discussion of project history profiling? Could go before the paragraph that starts with `Consider...`

stsievert: Ah, I see. I added a `\ref` command, which is what I normally use. When this document is compiled, it will produce "Figure 2" or something (and hyperlink the reference if we compile with that package).

scipy-1.0/paper.tex
```diff
@@ -83,23 +83,26 @@ \subsection*{Polynomial interpolators}
...
-    Consider the following asv benchmark results, spanning roughly nine years of project history, that demonstrate the gradual performance
-    improvements in \texttt{scipy.spatial.cKDTree.query()} (nearest-neighbor search) performance, using command:
+    Consider the asv benchmark results shown in Figure~\ref{fig:asvbench}, spanning roughly nine years of project history. These demonstrate the gradual performance
+    improvements in a nearest-neighbor search through \texttt{scipy.spatial.cKDTree.query()}, and can be run using the command:
+
 
     \texttt{python run.py run -e -s 800 --bench "\textbackslash btime\_query\textbackslash b" "02de46a546..b3ddb2c"}
 
-    \begin{figure}[H]
+    \begin{figure}
```
tylerjereddy: Removing the locator flag here lands Figure 1 in the numerical optimization section.

PR #24, opened 2018-04-12
ENH: optimize improvements: attempted to coordinate global and local …

mdhaber:
…minimization sections

### DIFF
--- a/scipy-1.0/scipy-optimize.tex
+++ b/scipy-1.0/scipy-optimize.tex
@@ -85,7 +85,7 @@ While most \texttt{minimize} development prior to SciPy 1.0 has focused on uncon
     iterate is sufficiently close to the solution: linear (L), superlinear (S) and quadratic (Q). Convergence rates denoted S$^*$ indicate that the algorithm
     has a superlinear rate for the parameters used in SciPy, but can  achieve a quadratic convergence rate with other parameter choices.
     \textit{Global convergence} is marked for the algorithms with guarantees of convergence to a stationary
-    point (i.e. a point $x^*$ for which $\nabla f(x^*) = 0$); this is \emph{not} a guarantee of convergence to a local or global minimum. The table also indicates which algorithms
+    point (i.e. a point $x^*$ for which $\nabla f(x^*) = 0$); this is \emph{not} a guarantee of convergence to a global minimum. The table also indicates which algorithms
     can deal with constraints on the variables. We distinguish between: \textit{bound constraints} (i.e. $x^l \le x \le x^u$),
     \textit{equality constraints} (i.e. $c_{\text{eq}}(x) = 0$) and \textit{inequality constraints} (i.e. $c_{\text{ineq}}(x) \ge 0$).}
   \begin{tabular}{cccccccccccccc}
@@ -127,7 +127,7 @@ While most \texttt{minimize} development prior to SciPy 1.0 has focused on uncon
 \end{table}
 
 \paragraph{Global Minimization}
-As \texttt{minimize} only seeks a local minimum, some problems require the use of a global optimization routine to find the minimum of the objective function over the set of all decision variable values. % , especially if the objective function possesses many local minima that other solvers could get stuck in.
+As \texttt{minimize} only seeks a local minimum, some problems require the use of a global optimization routine to find the minimum of the objective function over the set of all decision variable values, especially if the objective function possesses many local minima that other solvers could get stuck in. % e.g. if there is concern that `minimize` could get stuck in an unacceptable local minimum?
 The \texttt{brute} function performs a grid search over the entire domain, and the best grid point is ``polished'' by a local minimizer. However, this approach becomes prohibitively expensive as the problem dimensionality increases.
 % As far as I understand how brute works, the run time does not depend on the number of extrema, as it only polishes the best one
 % , or if the objective function has many extrema. 
--- a/scipy-1.0/scipy-optimize.tex
+++ b/scipy-1.0/scipy-optimize.tex
@@ -28,9 +28,9 @@ A presolve routine \cite{andersen1995presolving} solves trivial problems and oth
 Compared to the previously implemented simplex method, the new interior-point method is faster for all but the smallest problems, and is suitable for solving medium- and large-sized problems on which the existing simplex implementation fails. However, the interior point method typically returns a solution near the center of an optimal face, yet basic solutions are often preferred for sensitivity analysis and for use in mixed integer programming algorithms. This motivates the need for a crossover routine or a new implementation of the simplex method for sparse problems in a future release, either of which would require an improved sparse linear system solver with efficient support for rank-one updates
 
 \subsubsection*{Nonlinear Optimization}
+\paragraph{Local Minimization}
 
-
-The \texttt{minimize} function provides methods for solving nonlinear optimization
+The \texttt{minimize} function provides methods for finding local minima of nonlinear optimization
 problems. It unifies several methods with a common interface in order to facilitate
 interchanging between solvers with different characteristics, summarized in Table~\ref{tab:minimize}.
 %% Currently, this introduces a ton of whitespace around a very simple expression (min f(x)). The reader can easily find the standard of meaning of `nonlinear programming', so I believe it should be removed. If we decide that it is appropriate to include, then we should add the entire problem statement - including constraints - and we should also include the entire linear programming problem statement in the previous section.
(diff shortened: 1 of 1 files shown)


scipy-1.0/scipy-optimize.tex
```diff
@@ -83,7 +85,7 @@ \subsubsection*{Nonlinear Optimization}
     iterate is sufficiently close to the solution: linear (L), superlinear (S) and quadratic (Q). Convergence rates denoted S$^*$ indicate that the algorithm
     has a superlinear rate for the parameters used in SciPy, but can  achieve a quadratic convergence rate with other parameter choices.
     \textit{Global convergence} is marked for the algorithms with guarantees of convergence to a stationary
-    point (i.e. a point $x^*$ for which $\nabla f(x^*) = 0$). The table also indicates which algorithms
+    point (i.e. a point $x^*$ for which $\nabla f(x^*) = 0$); this is \emph{not} a guarantee of convergence to a local or global minimum. The table also indicates which algorithms
```
antonior92: I think it would be better to say ``this is \emph{not} a guarantee of convergence to a global minimum.``

The convergence toward a stationary point guarantee that the solver converges towards either a local minimum or a saddle point. So saying that it is not a guarantee of convergence toward a local minimum may not be wrong but may be misleading.

scipy-1.0/scipy-optimize.tex
```diff
@@ -124,9 +126,21 @@ \subsubsection*{Nonlinear Optimization}
...
-\texttt{Basinhopping} \cite{Wales1997} uses random perturbation of the parameter vector, with local minimization. The acceptance of the new coordinates is dictated by the Metropolis test typically used in Monte Carlo simulations.
-\texttt{differential\textunderscore evolution} \cite{Wormington1999,Storn1997} is a stochastic solver that improves by gradual evolution of a population of candidate solutions. In each iteration trial candidates are generated by combination of candidates from the existing population; if the trial candidates represent an improvement then the population is updated. \texttt{differential\textunderscore evolution} does not require that the objective function be differentiable.
+\paragraph{Global Minimization}
+As \texttt{minimize} only seeks a local minimum, some problems require the use of a global optimization routine to find the minimum of the objective function over the set of all decision variable values. % , especially if the objective function possesses many local minima that other solvers could get stuck in.
```
antonior92: I think you can remove the comment and show the hole phrase...

scipy-1.0/scipy-optimize.tex
```diff
@@ -28,9 +28,9 @@ \subsubsection*{Linear Optimization}
 Compared to the previously implemented simplex method, the new interior-point method is faster for all but the smallest problems, and is suitable for solving medium- and large-sized problems on which the existing simplex implementation fails. However, the interior point method typically returns a solution near the center of an optimal face, yet basic solutions are often preferred for sensitivity analysis and for use in mixed integer programming algorithms. This motivates the need for a crossover routine or a new implementation of the simplex method for sparse problems in a future release, either of which would require an improved sparse linear system solver with efficient support for rank-one updates
 
 \subsubsection*{Nonlinear Optimization}
+\paragraph{Local Minimization}
```
antonior92: Maybe ``Local Nonlinear Minimization``

antonior92:
Hi @mdhaber, thank you for doing this. I left a few comment for you to consider...

scipy-1.0/scipy-optimize.tex
mdhaber: A global optimizer is appropriate whenever 1) getting a global minimum or lower local minimum is important enough to justify additional computation and 2) the problem is not convex. It would say it is _especially_ needed if those two things apply and 3) there is particular reason to suspect that stationary points other than the global minimum will attract a local minimizer. The _number_ of local minima (beyond one) is only a part of (3). 

So I think the statement "especially if the objective function possesses many local minima that other solvers could get stuck in." is somewhat over-simplified for the fraction of the audience that doesn't already know why a global minimizer might be needed, and unnecessary for those who already know why a global minimizer might be needed.

scipy-1.0/scipy-optimize.tex
mdhaber: Do others second this?

mdhaber:
@antonior92 Thanks for the comments! Before I change things, would you take a look at my responses?

scipy-1.0/scipy-optimize.tex
antonior92: Basically because  while ``global convergence`` is not the same as convergence toward a local minimum, it is the next best thing.  If you have ``global convergence`` and the Hessian at the stationary point is positive definite than you have converged to the local minimum.

In some cases, you even have proof in the original paper of the convergence to a local minimum (and not just global convergence) and I have just said you have global convergence, for simplicity (e.g. trust-exact).

Hence when you say: ``this is \emph{not} a guarantee of convergence to a local or global minimum`` you are not wrong, but you put both statements in the same level. And while you have no ways at all of checking or guaranteeing global convergence, you can always check and, some time, even guarantee local convergence.

scipy-1.0/scipy-optimize.tex
antonior92: Matt I actually don't think it is an over simplification. I think the objective function getting stuck in bad local minima actually summarises 1), 2) and 3) in a more accurate way than saying these three points separately. That is:
1) is basically the same;
2) for non convex problems you don't actually need to use global solvers, so it is not a very good justification...
3) Seems to me the same thing as well...

scipy-1.0/scipy-optimize.tex
mdhaber: I agree those two should not be put at the same level. I just didn't want to mislead people into believing there is a guarantee of convergence to a local minimum. 

But you're right, it's more important to emphasize that they aren't guaranteed a global minimum. While a local minimum isn't necessarily guaranteed either, that's not as important to emphasize. The careful reader can infer that from what you wrote about convergence to a stationary point.

antonior92:
Matt, I think we can merge this when you fell ready...

scipy-1.0/scipy-optimize.tex
```diff
@@ -124,9 +126,21 @@ \subsubsection*{Nonlinear Optimization}
...
+\paragraph{Global Minimization}
+As \texttt{minimize} only seeks a local minimum, some problems require the use of a global optimization routine to find the minimum of the objective function over the set of all decision variable values, especially if the objective function possesses many local minima that other solvers could get stuck in. % e.g. if there is concern that `minimize` could get stuck in an unacceptable local minimum?
+The \texttt{brute} function performs a grid search over the entire domain, and the best grid point is ``polished'' by a local minimizer. However, this approach becomes prohibitively expensive as the problem dimensionality increases.
+% As far as I understand how brute works, the run time does not depend on the number of extrema, as it only polishes the best one
```
andyfaff: Sorry, what I wrote is probably not clear enough. Yes, `brute` only polishes the best one. What I meant was that the grid search increases as the power of N. Say there's 5 grid points per parameter.
1 params - 5 evals
2 params - 25 evals
3 params - 125 evals, etc.

scipy-1.0/scipy-optimize.tex
```diff
@@ -124,9 +126,21 @@ \subsubsection*{Nonlinear Optimization}
...
+% , or if the objective function has many extrema. 
+To address this, the
+\texttt{basinhopping} \cite{Wales1997} function uses random perturbation of the decision variables in conjunction with local minimization. The acceptance of the perturbed coordinates is dictated by the Metropolis test typically used in Monte Carlo simulations. 
+\texttt{differential\textunderscore evolution} \cite{Wormington1999,Storn1997} is a stochastic solver that improves by gradual evolution of a population of candidate solutions. In each iteration, trial candidates are generated by combination of candidates from the existing population; if the trial candidates represent an improvement then the population is updated. 
+% Is the following a unique feature of differential_evolution? If so, state it (although it seems that neither `brute`or `basinhopping` inherently require the a differentiable problem). If not, why mention it?
```
andyfaff: May not be worth mentioning it, but solvers that don't need to use gradients are valuable.

scipy-1.0/scipy-optimize.tex
```diff
@@ -124,9 +126,21 @@ \subsubsection*{Nonlinear Optimization}
...
-\texttt{Basinhopping} \cite{Wales1997} uses random perturbation of the parameter vector, with local minimization. The acceptance of the new coordinates is dictated by the Metropolis test typically used in Monte Carlo simulations.
-\texttt{differential\textunderscore evolution} \cite{Wormington1999,Storn1997} is a stochastic solver that improves by gradual evolution of a population of candidate solutions. In each iteration trial candidates are generated by combination of candidates from the existing population; if the trial candidates represent an improvement then the population is updated. \texttt{differential\textunderscore evolution} does not require that the objective function be differentiable.
+\paragraph{Global Minimization}
+As \texttt{minimize} only seeks a local minimum, some problems require the use of a global optimization routine to find the minimum of the objective function over the set of all decision variable values, especially if the objective function possesses many local minima that other solvers could get stuck in. % e.g. if there is concern that `minimize` could get stuck in an unacceptable local minimum?
```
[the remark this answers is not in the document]
mdhaber: If by 
> Matt, I liked the way you put it
you meant you liked
> % e.g. if there is concern that `minimize` could get stuck in an unacceptable local minimum.
then great. But I'm not sure, so I'll let you merge this PR as-is now and make this modification in your next PR.

scipy-1.0/scipy-optimize.tex
mdhaber: Certainly it may be worth mentioning, but I don't think it makes sense to single out only `differential_evolution` if the same is true of all the global optimizers.

If none of the global optimizers inherently require derivatives, then we should say "None of the global optimizers require that the objective function be differentiable." rather than "\texttt{differential\textunderscore evolution} does not require that the objective function be differentiable."

However, maybe your point is really that among the three options you believe that `differential_evolution` is the strongest when derivatives are unavailable. `brute` does not require derivatives to do the grid search, but perhaps it doesn't give a great answer unless a derivative-based local optimizer polishes the solution. `basinhopping` does not inherently require derivatives, but perhaps it performs best when it is used in conjunction with a derivative-based local optimizer. So is the claim that `differential_evolution` works quite well without reliance on a local optimizer that may require derivatives, and that is unique?

scipy-1.0/scipy-optimize.tex
mdhaber: OK. In the comment, I was just explaining why I removed the italicized portion:

> However, this approach becomes increasingly inefficient as the problem dimensionality increases, _or if the objective function has many extrema_."

because that part of the statement is not accurate, as the efficiency does not depend on the number of extrema.

mdhaber:
@antonior92 Great. The only thing is that you said:

> I liked the way you put it

but I wasn't sure whether you were refering to the commented-out wording I suggested or to the original wording. I'll let you merge this and fix that part separately, if need be.

mdhaber:
@andyfaff Thanks for the comments. Please do submit a separate PR if you think the bit about not requiring derivatives is worth including again in some form.

antonior92:
Ok, just merged it as it is.

> I wasn't sure whether you were refering to the commented-out wording I suggested or to the original wording.

I was referring to the original wording...

PR #25, opened 2018-04-16
ENH: update asv benchmarks with additional version of Cython

tylerjereddy:
Running the benchmarks with a third version of Cython (`0.18`) has almost completely closed the gap in the benchmark plot so this PR attempts to update the plot / text accordingly.


tylerjereddy:
The compiled PDF looks ok & this PR only has minor improvements to `asv` things that are already in the manuscript so self-merging to keep things going.

PR #27, opened 2018-05-22
ENH: add draft of test suite section.

tylerjereddy:
Added a small subsection covering the test suite; I'm probably missing some details, but it is a start.


tylerjereddy:
Ok, self-merging since this isn't code, we're short on paper-helper time, and there's an approving review.

tylerjereddy:
We could also move this information to i.e., a table in the SI if it eventually gets moved out of main manuscript.

PR #30, opened 2018-06-18
Drafted a subsection for LowLevelCallable

tylerjereddy:
First draft of a subsection about `LowLevelCallable`, which is mentioned in the proposed manuscript outline. Not my area of expertise, but just trying to keep things moving--feel free to request a substantial rewrite. I guess a code example might be more compelling here, but I'm not sure if we were really wanting to have code snippets in the ms.


tylerjereddy:
Self-merging the draft after 10 days without feedback -- refinements can be made later -- keeping the paper moving forward.

PR #31, opened 2018-06-27
ENH: add some authors and a title for manuscript

tylerjereddy:
I've drafted a title for the manuscript and added the steering committee members to the author list; I've also added the contributors to the paper so far if they weren't already included in the latter list. 

This somewhat circumvents the originally-described detailed procedure for authorship assignments, but hopefully isn't too controversial and helps us continue to move forward. Nothing in particular is meant by the current ordering of authors; I suppose people should start adding themselves / their affiliations when they next open a PR etc.

### DIFF
--- a/scipy-1.0/paper.tex
+++ b/scipy-1.0/paper.tex
@@ -14,7 +14,7 @@
 \newcommand{\xmark}{\ding{55}}%
 
 
-\title {SciPy 1.0 - Fundamental Algorithms for Scientific Computing in Python}
+\title{SciPy 1.0---Fundamental Algorithms for Scientific Computing in Python}
 
 \author[1]{Pauli Virtanen}
 \author[2,*]{Ralf Gommers}
--- a/scipy-1.0/paper.tex
+++ b/scipy-1.0/paper.tex
@@ -33,12 +33,11 @@
 \author[16]{Josh Wilson}
 \author[17]{Matthew Brett}
 \author[18]{Nikolay Mayorov}
-\author[19]{Tyler Reddy}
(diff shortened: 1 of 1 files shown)


tylerjereddy:
Suggestions from @jarrodmillman include: removing the version number & including the word "stable" in the title as well.

jarrodmillman:
I am late to the discussion, but this looks good to me.  If no one objects, I would recommend merging this tomorrow or so.  Then folks can add their names and affiliations directly.  I am happy to defer to others on who should be an author and in what order they should be listed.  But I don't see any reason to wait to add this information.  Ralf, Pauli, or someone else can always reorder the names and add anyone who is missing before we submit anything.

rgommers:
How about "Fundamental Algorithms for Scientific Computing in Python" for a title?

Not so keen on "stable", it's one of the few things 1.0 doesn't mean / didn't change. 1.0 isn't really more stable than 0.19.0

tylerjereddy:
I've adjusted the title based on the initial feedback here; I've not removed the `1.0` version number form the title since some association with the "major release" was intended for the paper, but I'm certainly open to removing that if there's a consensus / objection considering we are well into 1.x series now.

The author list probably doesn't even include all core devs -- I don't think there's a complete overlap between core devs / commit rights and steering committee members that are actually listed at the moment. Indeed, let's not put too much more energy into that if there's a solid plan in place to pull in the information.

tylerjereddy:
I'll leave this open for feedback for a few more days.

rgommers:
Thanks Tyler

PR #32, opened 2018-06-27
ENH: first draft of abstract for SciPy paper

tylerjereddy:
I drafted an abstract for the manuscript; feel free to criticize.

### DIFF
--- a/scipy-1.0/paper.tex
+++ b/scipy-1.0/paper.tex
@@ -28,7 +28,19 @@
 \keywords{Scientific computing, Python, Mathematics}
 
 \begin{abstract}
-TODO. Abstract must be under 200 words and not include subheadings or citations.
+SciPy is an open source scientific computing library for the Python programming language.
+SciPy 1.0 was released in late 2017, about 16 years after the original
+version 0.1 release. SciPy
+has become a \emph{de facto} standard in the community, with more than 600
+unique code contributors and millions of downloads per year, despite
+the absence of an official project publication, until now. The library
+includes functionality spanning clustering, Fourier transforms,
+integration, interpolation, file I/O, linear algebra, image processing,
+orthogonal distance regression, minimization algorithms, signal processing,
+sparse matrix handling, computational geometry, and statistics. In this
+work, we highlight some recent technical developments and provide
+an overview of the capabilities and development practices of the
+SciPy library.
 \end{abstract}
 \begin{document}
 
--- a/scipy-1.0/paper.tex
+++ b/scipy-1.0/paper.tex
@@ -27,20 +27,23 @@
 
 \keywords{Scientific computing, Python, Mathematics}
 
+% NOTE: some usage stats below are from https://libraries.io/pypi/scipy
+% as well as github metrics; normally one doesn't put citations directly
+% in abstract though (depends on field / journal)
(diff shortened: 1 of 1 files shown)


scipy-1.0/paper.tex
```diff
@@ -28,7 +28,19 @@
 \keywords{Scientific computing, Python, Mathematics}
 
 \begin{abstract}
-TODO. Abstract must be under 200 words and not include subheadings or citations.
+SciPy is an open source scientific computing library for the Python programming language.
+SciPy 1.0 was released in late 2017, about 16 years after the original
+version 0.1 release. SciPy
+has become a \emph{de facto} standard in the community, with more than 600
```
mdhaber: Given the broad scope of the journal, a reader might not understand which community this refers to - the Python community, or the scientific (computing) community? (Of course, the term SciPy also refers to the [SciPy ecosystem ](https://www.scipy.org/) as a whole (including matplotlib, sympy, pandas, etc...), so it could conceivably mean that community, too.) On the other hand, if you were to state explicitly "a _de facto_ standard for scientific computing in the Python community", this would seem redundant. So maybe it's fine as-is, but perhaps consider how it might be more explicit.

scipy-1.0/paper.tex
```diff
@@ -28,7 +28,19 @@
...
+unique code contributors and millions of downloads per year, despite
+the absence of an official project publication, until now. The library
```
mdhaber: Carefully considering [rules for comma usage](https://www.grammarly.com/blog/comma/), I can't find a rule that definitely matches this use before "until now". I'm not certain about this case, but usually (for me) this means that no comma is needed.

scipy-1.0/paper.tex
```diff
@@ -28,7 +28,19 @@
...
+includes functionality spanning clustering, Fourier transforms,
+integration, interpolation, file I/O, linear algebra, image processing,
+orthogonal distance regression, minimization algorithms, signal processing,
+sparse matrix handling, computational geometry, and statistics. In this
+work, we highlight some recent technical developments and provide
```
mdhaber: I typically refrain from making suggestions I can't back up with objective rationale... but for whatever reason, I feel strongly enough to suggest reversing the order of "overview of the capabilities and development practices..." and "highlight some recent technical developments". Perhaps it's for the same reason that we begin the paper with overviews and end with specifics. But if nobody else suggests the same, please ignore me : )

scipy-1.0/paper.tex
rgommers: How about adding number of dependent packages and repositories, as calculated by https://libraries.io/pypi/scipy?

scipy-1.0/paper.tex
rgommers: the "despite the absence ..." is not needed I think. that's more a letter to the editor kind of thing

rgommers:
reads well overall, thanks!

tylerjereddy:
I've attempted to address the first round of revisions--I'll leave this open for a few days to allow time for additional feedback.

tylerjereddy:
Ok, that's about a week without additional requests for revisions and two core devs were mostly +1 for content in their reviews after my changes so self-merging to keep moving paper forward.

PR #34, opened 2018-06-28
ENH: drafted roadmap section of manuscript

tylerjereddy:
I tried to draft a roadmap section for the manuscript, which is obviously based pretty strongly on our actual roadmap. It may suffer somewhat from the list-like structure of the original as far as being nicely-flowing prose goes.


scipy-1.0/paper.tex
```diff
@@ -185,6 +185,85 @@ \section*{Project organisation and community}
 external representative on the NumFOCUS Board of Directors.
 
 \textbf{Roadmap}
+The SciPy Roadmap\cite{SciPy_roadmap} is a continuously-updated document
+maintained by the community that describes some of the major directions
+for improvement for the project, as well as specific limitations and
+matters that require assistance moving forward.
+
+In general, we encourage changes that improvie clarity in the API of the library,
```
jarrodmillman: s/improvie/improve

scipy-1.0/paper.tex
```diff
@@ -185,6 +185,85 @@ \section*{Project organisation and community}
...
+including migration to typed memoryviews to handle NumPy arrays in place
+of the older syntax.
+
+Some specific submodules of SciPy would benefit from  attention moving forward.
+\texttt{fftpack} has some single-precision floating point errors, a caching bug,
+and would benefit from the addition of the Bluestein algorithm. There's also
+too much community code duplication with \texttt{numpy.fft}, which
+requires clarification of best choices and long-term plans.
+
+\texttt{scipy.integrate} now has a new ODE solver interface, and we should aim
```
jarrodmillman: s/now//

scipy-1.0/paper.tex
```diff
@@ -185,6 +185,85 @@ \section*{Project organisation and community}
...
+proving more effective in practice. We also need to standardize the
+\texttt{scipy.ndimage.morphology} API.
+
+\texttt{scipy.signal} needs its overlap with \texttt{scipy.ndimage}
+to be reduced; B-splines functionality should be moved to \texttt{interpolate},
+\texttt{firwin2} can be merged into \texttt{firwin}; Second Order Sections need
+to be updated to match the capabilities in other routines, and wavelet-related
+routines are somewhat out of place.
+
+\texttt{scipy.sparse} is fairly mature now, but there are many data
```
jarrodmillman: Maybe "While \texttt{scipy.sparse} is mature, there are ..."

scipy-1.0/paper.tex
```diff
@@ -185,6 +185,85 @@ \section*{Project organisation and community}
...
+
+For \texttt{scipy.linalg} we need to remove functions that are duplicate with 
+\texttt{numpy.linalg}, \texttt{get\_lapack\_funcs} should always use flapack, and
+we should continue to wrap more LAPACK functions. Ideally, we'd also
+add type-generic wrappers in the Cython BLAS and LAPACK interfaces, and
+transform many of the linear algebra routines into gufuncs.
+
+We're planning to remove \texttt{scipy.misc} as a public module, and
+rename \texttt{scipy.odr} to \texttt{scipy.regression} or
+\texttt{scipy.fitting}, and include \texttt{optimize.curve\_fit}. This
```
mdhaber: Consider "relocate..." or "migrate... to this new home" or similar. "include" might suggest that `optimize.curve\_fit` is not presently included in SciPy.

scipy-1.0/paper.tex
```diff
@@ -185,6 +185,85 @@ \section*{Project organisation and community}
...
+We're planning to remove \texttt{scipy.misc} as a public module, and
+rename \texttt{scipy.odr} to \texttt{scipy.regression} or
+\texttt{scipy.fitting}, and include \texttt{optimize.curve\_fit}. This 
+module will then provide a home for other fitting functionality.
+\texttt{scipy.optimize} would benefit from a few more good global optimizers 
+as well as large-scale optimizers.
+
+\texttt{scipy.ndimage} should be completely migrated to a data point
+model (values on a grid), instead of a pixel model (elements with centers),
+both of which have been used in the past, because the former is now
```
mdhaber: "the former" may be ambiguous here. Clearly you mean the first one you mentioned ("values on a grid", not necessarily the "former" choice of models temporally), but perhaps there is a better word choice given that you have just mentioned "the past".

tylerjereddy:
Rebased / force pushed and revised.

Things I'm not sure about:

1. SciPy tutorial page(s) traffic -- maybe a senior team member has access to that info if they also deem it pertinent here. For now, I just indicated that there are 15 sections to our tutorial along with a citation to the online resource.
2. For the typed memoryview clarification requested I added a short code listing / snippet -- as I recall there was some informal objection to code listings in the ms. during the planning stages so this may require some justification or just switching to an inline prose explanation somehow.

scipy-1.0/paper.tex
```diff
@@ -185,6 +186,114 @@ \section*{Project organisation and community}
...
+\end{figure}
+
+The low-level Cython code in our library (which interacts with C-level
+code and exposes it for Python usage) could use some measure of modernization,
+including migration to typed memoryviews to handle NumPy arrays instead of
+the older syntax:
+
+\begin{lstlisting}[language=Python]
+# example of a function that takes a single
+# 3 dimesional typed memoryview as an argument
```
endolith: "dimensional" typo

rgommers:
>  It may suffer somewhat from the list-like structure of the original as far as being nicely-flowing prose goes.

I think it would be good to limit the discussion to key future changes, and remove the rest. E.g. `fftpack`, `io`, `KDTree` are all not very interesting to a reader. Of the ones mentioned, these are the interesting entries I think: `interpolate`, `linalg`, `optimize`, `ndimage`, `signal`.

`stats` is missing. The addition of *sparse arrays* is also worth mentioning.

scipy-1.0/paper.tex
```diff
@@ -185,6 +186,114 @@ \section*{Project organisation and community}
...
+too much community code duplication with \texttt{numpy.fft}, which
+requires clarification of best choices and long-term plans.
+
+\texttt{scipy.integrate} has a new ODE solver interface, and we should aim
+to gracefully deprecate the older API. Support for integrating complex-valued
+functions and integrating multiple intervals could be beneficial.
+
+\texttt{scipy.interpolate} could benefit from a set of new features including
+spline fitting routines with better user control; integration, differentiation,
+and arithmetic routines for splines; transparent tensor-product splines,
```
ev-br: Not sure what integration and differentiation means here: this has been implemented. (The rest I totally agree with.)

scipy-1.0/paper.tex
```diff
@@ -185,6 +186,114 @@ \section*{Project organisation and community}
...
 
 \textbf{Roadmap}
+The SciPy Roadmap\cite{SciPy_roadmap} is a continuously-updated document
+maintained by the community that describes some of the major directions
+for improvement for the project, as well as specific limitations and
+matters that require assistance moving forward.
+
+In general, we encourage changes that improve clarity in the API of the library,
+but strongly discourage breaking backwards compatibility, given our position near
+the base of the scientific Python computing stack.
```
rgommers: Good paragraph, but let's move it to the API/ABI evolution section

scipy-1.0/paper.tex
```diff
@@ -185,6 +186,114 @@ \section*{Project organisation and community}
...
+
+Unit test coverage for most recent code additions / adjustments is quite
+respectable, with overall project test coverage holding steady at
+around 76\% for the last six months (Figure~\ref{fig:coverage}),
+but some of the historical components of the code base
+would still benefit from increased test coverage. We are still striving
```
rgommers: This can go to the testing/benchmarking section

scipy-1.0/paper.tex
```diff
@@ -185,6 +186,114 @@ \section*{Project organisation and community}
...
+to increase the number of high-quality SciPy usage tutorials beyond
```
rgommers: probably should remove "high-quality", because that's not really what they are I think - they could use some work

scipy-1.0/paper.tex
```diff
@@ -185,6 +186,114 @@ \section*{Project organisation and community}
...
+\begin{figure}[H]
+\centering
+\includegraphics[width=\textwidth]{static/coverage-chart}
+\caption{\% coverage of the SciPy code base by unit tests over the six
+	 months preceding June 28, 2018 as reported by the \texttt{codecov}
+	 service. The analysis of lines of source code covered is performed
+	 automatically in our continuous integration suite using the
+	 \texttt{pytest-cov} and \texttt{gcov} libraries.}
+\label{fig:coverage}
+\end{figure}
```
rgommers: nice figure, let's keep that in the testing section as well

scipy-1.0/paper.tex
```diff
@@ -185,6 +186,114 @@ \section*{Project organisation and community}
...
+\end{lstlisting}
+
+\texttt{fftpack} has some single-precision floating point errors, a caching bug,
+and would benefit from the addition of the Bluestein algorithm. There's also
+too much community code duplication with \texttt{numpy.fft}, which
+requires clarification of best choices and long-term plans.
+
+\texttt{scipy.integrate} has a new ODE solver interface, and we should aim
+to gracefully deprecate the older API. Support for integrating complex-valued
+functions and integrating multiple intervals could be beneficial.
```
rgommers: I'd remove the bit about the old API for sure. The rest is probably also not major enough to mention

scipy-1.0/paper.tex
```diff
@@ -185,6 +186,114 @@ \section*{Project organisation and community}
...
+# the same scenario
+cpdef int func_old(object[int, ndim=3, mode='strided'] arr):
+    pass
+
+\end{lstlisting}
+
+\texttt{fftpack} has some single-precision floating point errors, a caching bug,
+and would benefit from the addition of the Bluestein algorithm. There's also
+too much community code duplication with \texttt{numpy.fft}, which
+requires clarification of best choices and long-term plans.
```
rgommers: I propose to remove the first bit, and mention the overlap with numpy for both `fftpack` and `linalg` together

scipy-1.0/paper.tex
```diff
@@ -185,6 +186,114 @@ \section*{Project organisation and community}
...
+
+\texttt{scipy.interpolate} could benefit from a set of new features including
+spline fitting routines with better user control; integration, differentiation,
+and arithmetic routines for splines; transparent tensor-product splines,
+Non-Uniform Rational B-Splines support; and mesh refinement and coarsening of
+B-splines and corresponding tensor products.
+
+\texttt{scipy.io} would benefit from PCM float support, but we otherwise
+direct users with more sophisticated needs to domain-specific audio
+libraries.
```
rgommers: let's remove this, too minor

scipy-1.0/paper.tex
```diff
@@ -185,6 +186,114 @@ \section*{Project organisation and community}
...
+
+For \texttt{scipy.linalg}, we need to remove functions that are duplicate with 
+\texttt{numpy.linalg} unless performance is superior in SciPy,
+\texttt{get\_lapack\_funcs} should always use flapack, and
+we should continue to wrap more LAPACK functions. Ideally, we'd also
+add type-generic wrappers in the Cython BLAS and LAPACK interfaces, and
+transform many of the linear algebra routines into gufuncs.
```
rgommers: let's remove the `flapack` stuff, not very user relevant. The type-generic wrappers thing we could keep here, but only if we rephrase it so that it makes sense to the average reader

scipy-1.0/paper.tex
```diff
@@ -185,6 +186,114 @@ \section*{Project organisation and community}
...
+
+We're planning to remove \texttt{scipy.misc} as a public module, and
+rename \texttt{scipy.odr} to \texttt{scipy.regression} or
+\texttt{scipy.fitting}, and relocate \texttt{optimize.curve\_fit}. This 
+module will then provide a home for other fitting functionality.
+\texttt{scipy.optimize} would benefit from a few more good global optimizers 
+as well as large-scale optimizers.
```
rgommers: I propose just to keep the `optimize` part. E.g.: The `optimize` module will keep evolving, we expect to add several new global optimization methods with improved performance over the existing ones, as well as add large-scale optimizers.

scipy-1.0/paper.tex
```diff
@@ -185,6 +186,114 @@ \section*{Project organisation and community}
...
+
+\texttt{scipy.ndimage} should be completely migrated to a data point
+model (values on a grid), instead of a pixel model (elements with centers),
+both of which have been used in the past, because the data point model is now
+proving more effective in practice. We also need to standardize the
+\texttt{scipy.ndimage.morphology} API.
```
rgommers: let's keep the pixel-vs-point here, and remove the morphology bit. @stefanv has a draft section on pixel-vs-point, we can align these two sections later after merging

scipy-1.0/paper.tex
```diff
@@ -185,6 +186,114 @@ \section*{Project organisation and community}
...
+
+\texttt{scipy.signal} needs its overlap with \texttt{scipy.ndimage}
+to be reduced; B-splines functionality should be moved to \texttt{interpolate},
+\texttt{firwin2} can be merged into \texttt{firwin}; Second Order Sections need
+to be updated to match the capabilities in other routines, and wavelet-related
+routines are somewhat out of place.
```
rgommers: I suggest to mention `signal` in the section that's now about `interpolate` (talk about all spline topics there). SOS is a good topic perhaps. I would remove the rest.

scipy-1.0/paper.tex
```diff
@@ -185,6 +186,114 @@ \section*{Project organisation and community}
...
+
+While \texttt{scipy.sparse} is mature, there are many data
+formats and the community continues to move forward here so we will aim
+to take advantage of their improvements while preserving backwards
+compatibility to the extent that it is possible. \texttt{sparse.linalg}
+may benefit from wrappers for PROPACK for faster sparse SVD computation.
```
rgommers: Two things I'd mention here: sparse arrays (https://github.com/pydata/sparse) and (perhaps) PROPACK.

scipy-1.0/paper.tex
```diff
@@ -185,6 +186,114 @@ \section*{Project organisation and community}
...
+
+For \texttt{scipy.spatial}, KDTree will be removed, and cKDTree will be 
+renamed to KDTree in a backwards-compatible way. \texttt{distance\_wrap.c}
```
rgommers: I wouldn't mention the `KDTree` removal, users won't notice since it becomes an alias of `cKDTree`. There's nothing in `spatial` important enough to mention AFAIK.

scipy-1.0/paper.tex
```diff
@@ -185,6 +186,114 @@ \section*{Project organisation and community}
...
+formats and the community continues to move forward here so we will aim
+to take advantage of their improvements while preserving backwards
+compatibility to the extent that it is possible. \texttt{sparse.linalg}
+may benefit from wrappers for PROPACK for faster sparse SVD computation.
+
+For \texttt{scipy.spatial}, KDTree will be removed, and cKDTree will be 
+renamed to KDTree in a backwards-compatible way. \texttt{distance\_wrap.c} 
+likely needs a rewrite in Cython. \texttt{scipy.special} needs precision
+improvements for hypergeometric functions, parabolic cylinder functions,
+and spheroidal wave functions.
```
rgommers: we may keep the `special` stuff, I think it's a big topic.

scipy-1.0/paper.tex
rgommers: Sorry changed my mind: for `spatial` we can mention `spatial.transforms` with support for rotation matrices.

scipy-1.0/paper.tex
rgommers: The most important thing for `linalg` I think is to start upgrading to more recent LAPACK versions.

ilayn: I think for the average reader it is more important to mention that they can access the LAPACK funcs via either `get_lapack_funcs` or via `sp.linalg.lapack.<whatever>`.

ilayn:
You might want to add some more options for the code listings to render them a bit prettier, this is the current rendering 

<hr>

image
<hr>
If you also add `textcomp` package and add the following 

```latex
\usepackage{listings, textcomp}
    
\lstset{ %
  basicstyle=\ttfamily\footnotesize,  % size of fonts used for the code
  breaklines=true,   % automatic line breaking only at whitespace
  captionpos=b,   % sets the caption-position to bottom
  commentstyle=\color{gray},  % comment style
  keywordstyle=\color{blue},  % keyword style
  stringstyle=\color{red},  % string literal style
  upquote=true  %straight single quotes (requires textcomp)
}
```
it becomes 

image


I am not suggesting that this is the optimal output but in my opinion it is a bit better in terms of blending into text. Maybe a frame or enabling line numbers can make it more obvious .

tylerjereddy:
I tried to address most of the revisions, but noticed I forgot the typo pointed out by @endolith and the recent code format thing from @ilayn.

The revisions were pretty major & rebasing a little messy so good to double check.

tylerjereddy:
Ok, I think I've addressed those two points now as well & added @ilayn as co-author on latest commit since I copy-pasted the suggested  adjustment to code listing appearance.

rgommers:
LGTM now, merged. Thanks Tyler

ilayn:
Just for the record, you can skip the coauthor step :) My contribution is really minor stylistic stuff

PR #36, opened 2018-06-29
[WIP] draft of language choices subsection

tylerjereddy:
Just a short paragraph and a plot of the language breakdown for the SciPy library for now. I'm not sure if this is the kind of thing actually intended for this section.

Some thoughts:

- would be helpful to have the Python / Cython languages separated -- is this sufficiently useful to justify i.e., opening a PR to the linguist library so they can distinguish those languages, or do we have a better way to parse this anyway?
- the plot is probably too big and even a table or mention in the text might suffice I suppose
- could perhaps draft a little bit about some of the libraries we wrap over in i.e., Fortran?


ilayn:
I think I am developing an OCD for images with nonmatching fonts 😃 . Maybe I should do the whole image suite of the article once we are done with the content.  But here it is anyways

```latex
\documentclass[fleqn,10pt]{wlscirep}
\usepackage{pgfplots}% Loads TikZ implicitly
\pgfplotsset{compat=newest}
\begin{document}
\begin{figure}\centering%
\begin{tikzpicture}[font=\sffamily]
\begin{axis}[height=6cm, width=8cm,
ybar, ymax=60, ymin=0,bar width=1cm,
ylabel={\% of Scipy code base},
ytick={0,20,...,60},
axis lines*=left,
axis y line shift=10pt,
enlarge x limits=0.15,
symbolic x coords={py,f,c,cpp},
xticklabels={Python \\\& Cython,FORTRAN,C,C++},
ticklabel style={align=center,font=\sffamily\small,
/pgf/number format/assume math mode=true},
xtick=data,
nodes near coords,
nodes near coords style={/pgf/number format/assume math mode=true}
]
\addplot coordinates {(py,52.4) (f,25.6) (c,19.5) (cpp,2.3)};
\end{axis}
\end{tikzpicture}
\end{figure}
\end{document}
```

image

scipy-1.0/paper.tex
```diff
@@ -54,7 +54,24 @@ \subsection*{Submodule organisation}
 
 \subsection*{Common infrastructure}
 
-\subsection*{Language preferences}
+\subsection*{Language choices}
+Python / Cython, Fortran, C and C++ are the programming languages
+used to implement scientific algorithms in the SciPy library. An analysis
+of our code base using the \texttt{linguist} library provides a detailed
```
ilayn: It is also written on top of the GitHub repo (click the color bar), where I got the numbers from.

rgommers:
+1 for splitting Python and Cython. That's kind of essential for what I planned to write on this. Which is:
- Our preference of implementation languages, in order.
- Reasons why that's our preference.

tylerjereddy:
Ok, I've used a feature branch of linguist to generate the language stats separately for Cython and Python now. [I've opened a PR with them](https://github.com/github/linguist/pull/4181) -- since that's an official github repo for `linguist` it would actually update the language stats at the top of all github project bars to distinguish the two languages, but only if they agree that splitting Cython off is a good idea. Regardless, we have an estimate for the paper now.

I've incorporated the Tikz template provided by @ilayn, modified it to reflect the new Cython-split language stats, and added @ilayn as co-author on the revision commit.

We'll still need more content for this section, but this is a step forward anyway I think.

tylerjereddy:
I've substantially expanded the discussion of the languages used in SciPy and some of the libraries that are wrapped from those languages, along with quite a large number of new citations to accompany them.

This is probably ready for another round of review now. It is unlikely that the content I've drafted is completely historically accurate but the project veterans will probably be happy to comment on that.  This may also be more detail than needed, but could also just make a table in the supporting info for the wrapped libraries if the paper starts getting too long.

tylerjereddy:
Also, rebased & force-pushed on latest master so the CI compiled PDF is our latest version of paper.

scipy-1.0/paper.tex
```diff
@@ -113,7 +116,94 @@ \subsection*{Submodule organisation}
...
+is an extremely well-established scientific programming language, both
+for historical reasons and because of its continued excellent
+performance\cite{Koelbel:1993:HPF:562354}. We
+wrap the FFTPACK Fortran library for performing Fourier
+transforms\cite{SWARZTRAUBER198445, SWARZTRAUBER198251} since
+this library has been a standard in the field for 33 years and has a license
+that is compatible with our own. Likewise, we wrap the Fortran source
+for ODEPACK\cite{citeulike:2644528} as it has been  trusted for the initial 
+value problem for ordinary differential equation systems for 30 years. For
+similar reasons we also warp the Fortran libraries QUADPACK\cite{1983qspa.book.....P} (for numerical
```
rgommers: typo: "warp" --> "wrap"

PR #37, opened 2018-07-07
Add section on stats distributions

tylerjereddy:
Though I'm (really) not the person to draft the stats 'recent technical improvements' section, sometimes it is helpful to have a starting skeleton for the experts to chip away at / improve (or direct me to make changes).

My strategy for compiling this section, since I only occasionally review stats PRs, was to parse the changelogs for scipy.stats over the last few releases and just look for i.e., major additions or design changes.

For a start I tried to parse the logs back through 0.17, but not 0.16 (yet--depending on how much we want in here).

### DIFF
--- a/scipy-1.0/paper.tex
+++ b/scipy-1.0/paper.tex
@@ -264,28 +264,29 @@ from a Cython module using \texttt{scipy.LowLevelCallable.from\_cython}.
 
 \subsection*{Statistical distributions}
 Version 1.0 of SciPy saw the addition of the energy distance algorithm in
-\texttt{scipy.stats.energy\_distance}. Energy distance is used in a varity of applications ranging
-from genetic analysis to multivariate independence. We also added the
+\texttt{scipy.stats.energy\_distance}. Energy distance is used in a variety
+of applications ranging from genetic analysis\cite{Hu2008DetectingIC} to 
+multivariate independence\cite{szekely2009}. We also added the
 first Wasserstein distance algorithm to the library in
 \texttt{scipy.stats.wasserstein\_distance}. The Wasserstein distance is often
-used in color digital image analysis.
+used in color digital image analysis \cite{192468}.
 
-We now also provide the cumulative distribution function of the multivariate
+We now provide the cumulative distribution function of the multivariate
 normal distribution, \\\texttt{scipy.stats.multivariate\_normal}, via the newly
-associated \texttt{logcdf} and \texttt{cdf} methods.
+associated \texttt{logcdf} and \texttt{cdf} methods. Other recent distributions 
+added to \texttt{scipy.stats} include the Argus distribution
+in \texttt{scipy.stats.argus} (used in particle physics\cite{ALBRECHT1990278}), a histogram-based
+distribution in \texttt{scipy.stats.rv\_histogram}, the multinomial
+distribution in \texttt{scipy.stats.multinomial} (used, for example, in natural
+language processing\cite{Griffiths5228}), the Burr type XII distribution in
+\texttt{scipy.stats.burr12}, the skew normal distribution 
+in \texttt{scipy.stats.skewnorm}, the trapezoidal distribution in 
+\texttt{scipy.stats.trapz}, and the matrix normal distribution
+in \texttt{scipy.stats.matrix\_normal}.
 
-Other recent additions to \texttt{scipy.stats} include the Argus distribution
-in \texttt{scipy.stats.argus} (used in particle physics), a class for
-generating a distribution from a histogram in
-\texttt{scipy.stats.rv\_histogram}, an implementation of the multinomial
-distribution in \texttt{scipy.stats.multinomial} (used for example in natural
-language processing), a weighted Kendall's tau statistic algorithm in
-\texttt{scipy.stats.weightedtau}, interquantile range handling in
-\texttt{scipy.stats.iqr}, the Burr type XII distribution is available in
-\texttt{scipy.stats.burr12}, the skew normal distribution has been implemented 
-as \texttt{scipy.stats.skewnorm}, a trepzoidal continuous random variable
-has been added in \texttt{scipy.stats.trapz}, and the matrix normal distribution
-is now available in \texttt{scipy.stats.matrix\_normal}. 
+There's also a new weighted Kendall's tau statistic algorithm in
+\texttt{scipy.stats.weightedtau}, and interquantile range handling in
+\texttt{scipy.stats.iqr} 
 
 From a design standpoint, many functions in \texttt{scipy.stats} now support
 a \texttt{nan\_policy} keyword to allow user-level control over the algorithmic
--- a/scipy-1.0/paper.tex
+++ b/scipy-1.0/paper.tex
@@ -263,14 +263,16 @@ from a Cython module using \texttt{scipy.LowLevelCallable.from\_cython}.
 
 
 \subsection*{Statistical distributions}
-The \texttt{scipy.stats} module provides a framework to implement probability
-distributions.  Given an expression for a probability distribution function (pdf),
+The \texttt{scipy.stats} module provides provides more than 100 probability
+distributions and a framework to implement additional distributions. Given
+an expression for a probability distribution function (pdf),
 the framework will calculate moments, cumulative distribution function (cdf),
 differential entropy, and more. Those properties are calculated based on
-generic code, e.g., numerical integration of the pdf to obtain the cdf. In cases
-where that generic code does not provide sufficient accuracy, it is replaced
-by an implemented specific to the distribution.
-The distributions framework also allows the user to draw random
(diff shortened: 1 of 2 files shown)


scipy-1.0/paper.tex
```diff
@@ -151,6 +151,34 @@ \subsection*{Numerical optimization}
...
+from genetic analysis to multivariate independence. We also added the
+first Wasserstein distance algorithm to the library in
+\texttt{scipy.stats.wasserstein\_distance}. The Wasserstein distance is often
+used in color digital image analysis.
+
+We now also provide the cumulative distribution function of the multivariate
+normal distribution, \\\texttt{scipy.stats.multivariate\_normal}, via the newly
+associated \texttt{logcdf} and \texttt{cdf} methods.
+
+Other recent additions to \texttt{scipy.stats} include the Argus distribution
```
mdhaber: Maybe all the distributions from this paragraph could be moved to the previous paragraph (so there is one paragraph that just lists all the new distributions)?

scipy-1.0/paper.tex
```diff
@@ -151,6 +151,34 @@ \subsection*{Numerical optimization}
...
+in \texttt{scipy.stats.argus} (used in particle physics), a class for
+generating a distribution from a histogram in
+\texttt{scipy.stats.rv\_histogram}, an implementation of the multinomial
+distribution in \texttt{scipy.stats.multinomial} (used for example in natural
+language processing), a weighted Kendall's tau statistic algorithm in
+\texttt{scipy.stats.weightedtau}, interquantile range handling in
+\texttt{scipy.stats.iqr}, the Burr type XII distribution is available in
```
mdhaber: Everything before Burr is written like "____ in _____", then the structure starts varying like "_____ is available in _____" and "_____ has been implemented as _____". I would suggest staying consistent throughout or changing every time.

scipy-1.0/paper.tex
```diff
@@ -151,6 +151,34 @@ \subsection*{Numerical optimization}
...
+
+We now also provide the cumulative distribution function of the multivariate
+normal distribution, \\\texttt{scipy.stats.multivariate\_normal}, via the newly
+associated \texttt{logcdf} and \texttt{cdf} methods.
+
+Other recent additions to \texttt{scipy.stats} include the Argus distribution
+in \texttt{scipy.stats.argus} (used in particle physics), a class for
+generating a distribution from a histogram in
+\texttt{scipy.stats.rv\_histogram}, an implementation of the multinomial
+distribution in \texttt{scipy.stats.multinomial} (used for example in natural
```
mdhaber: I like that you've given an example use of the multinomial distribution. It would be nice if the stats experts could include similar commentary for other new features - or it would be OK if we decide not to give examples uses at all. So I suppose the suggestion is to be consistent in the level of detail provided for these items that are listed like this unless there is a particular reason to highlight one as you have.

If this example were to stand alone, I believe "for example" deserves to be surrounded by parentheses here:
"used, for example, in natural language processing"
or maybe used like:
"used in natural language processing, for example"
https://data.grammarbook.com/blog/commas/commas-part-10/

tylerjereddy:
Thanks @mdhaber, I tried to address your revisions. For now, I've also added citations for the applications / use cases, but as you say we may very well end up removing those if they aren't applied consistently.

tylerjereddy:
Also, rebase / force push

scipy-1.0/paper.tex
```diff
@@ -249,6 +249,35 @@ \subsection*{Numerical optimization}
...
+\texttt{scipy.stats.energy\_distance}. Energy distance is used in a variety
+of applications ranging from genetic analysis\cite{Hu2008DetectingIC} to 
+multivariate independence\cite{szekely2009}. We also added the
+first Wasserstein distance algorithm to the library in
+\texttt{scipy.stats.wasserstein\_distance}. The Wasserstein distance is often
+used in color digital image analysis \cite{192468}.
+
+We now provide the cumulative distribution function of the multivariate
+normal distribution, \\\texttt{scipy.stats.multivariate\_normal}, via the newly
+associated \texttt{logcdf} and \texttt{cdf} methods. Other recent distributions
```
ev-br: Maybe mention that cdf uses under the hood the Fortran code by Alan Genz (scipy/stats/mvndst.f) which probably everybody else is using.

scipy-1.0/paper.tex
```diff
@@ -249,6 +249,35 @@ \subsection*{Numerical optimization}
...
+\texttt{scipy.stats.wasserstein\_distance}. The Wasserstein distance is often
+used in color digital image analysis \cite{192468}.
+
+We now provide the cumulative distribution function of the multivariate
+normal distribution, \\\texttt{scipy.stats.multivariate\_normal}, via the newly
+associated \texttt{logcdf} and \texttt{cdf} methods. Other recent distributions 
+added to \texttt{scipy.stats} include the Argus distribution
+in \texttt{scipy.stats.argus} (used in particle physics\cite{ALBRECHT1990278}), a histogram-based
+distribution in \texttt{scipy.stats.rv\_histogram}, the multinomial
+distribution in \texttt{scipy.stats.multinomial} (used, for example, in natural
```
ev-br: This reads like a long list from the release notes. Maybe better make `rv_histogram` stand out since it's actually a means for constructing a family of distributions? 
Ditto for `multinomial`, which is the first and so far the only multivariate discrete distribution.

rgommers: let's just delete the rest I'd say, something like `Argus` is not really big enough to mention in this paper imho

rgommers:
This is a bit different from what I had in mind. I was just thinking to describe the (improvements to) the distributions framework, and then highlight a couple of the interesting distributions (e.g. `histogram` and `multinomial`).

The `nan_policy` paragraph fits in well though.

jarrodmillman:
It would be nice to say something about ``random_state`` (with an example).

tylerjereddy:
Revised based on feedback in this PR and in-person discussion; hopefully condensed to more key points now. I think the plan is to have @rgommers refine at this stage.

Also, rebased/ force pushed.

rgommers:
@ev-br or @WarrenWeckesser does one of you have time for a quick review of this one?

scipy-1.0/paper.tex
```diff
@@ -263,6 +263,26 @@ \subsection*{Numerical optimization}
 
 
 \subsection*{Statistical distributions}
+The \texttt{scipy.stats} module provides a framework to implement probability
+distributions.  Given an expression for a probability distribution function (pdf),
+the framework will calculate moments, cumulative distribution function (cdf),
+differential entropy, and more. Those properties are calculated based on
+generic code, e.g., numerical integration of the pdf to obtain the cdf. In cases
```
WarrenWeckesser: I would replace the sentence

> In cases where that generic code does not provide sufficient accuracy, it is replaced by an implemented specific to the distribution.

with something like

> Where possible, the generic code is replaced by explicit formulas or more accurate or efficient implementations.  For example, most distributions have an explicit formula for the cdf, so it is used instead of integrating the pdf.

WarrenWeckesser:
Looks good.  I made one suggestion in-line.

Why are all those references being added in this pull request?  Will they be referenced in the text eventually?  Currently the text in this PR only references Griffiths5228.

scipy-1.0/paper.tex
```diff
@@ -263,6 +263,26 @@ \subsection*{Numerical optimization}
 
 
 \subsection*{Statistical distributions}
+The \texttt{scipy.stats} module provides a framework to implement probability
+distributions.  Given an expression for a probability distribution function (pdf),
```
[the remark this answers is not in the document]
rgommers: @mdhaber both. Users can implement their own, but it's not that common because so many distributions are already built-in.

@ev-br thanks, I'll take that over.

scipy-1.0/paper.tex
rgommers: thanks, that reads nicer

rgommers:
> Why are all those references being added in this pull request? Will they be referenced in the text eventually? Currently the text in this PR only references Griffiths5228.

They were in the PR already before we rewrote it to focus just on distributions. I'll take them out.

rgommers:
> One further small suggestion, in responce to Matt's comment. random_state might possibly get a code example, depending on a general policy of whether we include them or not.

I thought that we'd try to avoid code examples; definitely no long ones (working code), at most 1-4 lines to highlight things. That could be done here, but I think I'd prefer not to.

rgommers:
PR updated for all comments

ev-br:
Ok, hit the green button: I think it's better to have content in, and then we can iterate on it. And we'll inevitably run through the text as a whole. 
Thanks all.

chrisb83:
I just saw the discussion on stats here after i openend issue #97 that contains some ideas. I will read thorugh the above soon

PR #45, opened 2018-07-25
Fuse subsections for kdtree tech improvements; add a few refs

tylerjereddy:
Fusing the subsections in kdtree tech improvements section was briefly discussed in the BIDS sprint; also tried to add one or two more refs that might be relevant.

PR #49, opened 2018-07-25
Fix headings.

hameerabbasi:
The headings were being displayed as `0.0.1`, ...

This fixes that.


rgommers:
gh-45 already gets rid of those subheadings, so no need for this. Thanks though

PR #50, opened 2018-07-25
Add a script to count the number of public objects in SciPy submodules.

rgommers:
This is not perfect (e.g. doesn't count the few things in the main `scipy`
namespace, probably does a bit of double counting of stuff), but it should
be good enough to give an estimate.

If someone has a better script lying around, feel free to submit it or point me to it.

### DIFF
--- a/scipy-1.0/supporting_info/measure_scipy_API_surface.py
+++ b/scipy-1.0/supporting_info/measure_scipy_API_surface.py
@@ -1,0 +1,84 @@
+"""
+This script imports all the public submodules and estimates the number
+of public objects in each of them.  This can be used to give a number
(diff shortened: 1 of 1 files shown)


ev-br:
`refguide_check` does something similar, it IIRC looks into the `__all__` dicts. 
Just a quick-and-dirty check, https://github.com/scipy/scipy/compare/master...ev-br:api_surface?expand=1
shows `total =  1539 deprecated =  0 other =  11` for some random not-too-old dev version.
So the order of magnitude is about the same.

rgommers:
Thanks @ev-br, totally forgot about `refguide_check`. That does a much better job at not double-counting; your number is more accurate. This PR is then not too useful; better to add this to `refguide_check` as an option. Closing.

PR #52, opened 2018-07-25
Improvements to the Impact Now section

tylerjereddy:
* added a more exact count of SciPy PyPI downloads from the year 2017,
along with a citation for the software library we used to obtain
the metric

* removed the previous examples of prominent usages of SciPy and replaced
with more compelling examples including: LIGO work and Apple / Intel shipping
SciPy in distributions


tylerjereddy:
Still need some more compelling examples though, this is likely just a start.

rgommers:
needs a rebase already:)

tylerjereddy:
rebase / force push

rgommers:
LGTM, merged. Thanks @tylerjereddy

scipy-1.0/paper.tex
```diff
@@ -461,19 +461,16 @@ \section*{Discussion}
...
+From the user side, there were 13,096,468 PyPI downloads
+of SciPy during the year 2017\cite{pypinfo}, which is likely a vast
+underestimate of total downloads given that we don't have access to the
+popular Anaconda package manager download statistics. The recommended 
+source for citing SciPy is simply a reference to the website / library
+itself\cite{SciPylib}, and has been cited $>3000$ times. Some of the most
+prominent usages of or demonstrations of credibility for SciPy include the LIGO / Virgo
+scientific collaboration that lead to the observation of gravitational
+waves \cite{PhysRevLett.116.061102}, and the fact that SciPy is shipped directly
+with MacOS and in the Intel Distribution for Python\cite{intel-python}.
```
mdhaber: These are really strong!

For those (few) who are not already aware that SciPy is part of the Anaconda distribution of Python, the part earlier about not having access to the Anaconda download statistics might be confusing. 

For those who are aware, it might seem missing from this list at the bottom.

rgommers: Thoughts here: this is for a general journal, so mostly non-users. Intel is a name anyone recognizes, while Anaconda is not. Hence the choice for Intel. People who do use Python will already know that Anaconda is the most popular distribution

mdhaber: Agreed... but these are precisely the reasons I thought changes might be appropriate. Just to clarify my reasoning:

"this is for a general journal, so mostly non-users"
This is why I thought it might be confusing above when it says "we don't have access to the popular Anaconda package manager download statistics". I understand that non-users are supposed to infer from this that Anaconda is a very common way for people to install SciPy, I'm just not sure that they will.

"People who do use Python will already know that Anaconda is the most popular distribution" 
This is why I thought that the fact that SciPy is included in Anaconda is such a good "demonstration of credibility", especially with Anaconda being mentioned earlier. Both Intel and Anaconda could be listed here,

rgommers: how about we just rephrase the download statistics part as "underestimate of total downloads given that downloading from PyPI is only one of several popular methods to install SciPy".

mdhaber: Sure

rgommers: done in gh-59

PR #54, opened 2018-07-26
Initial Draft of Section on Cython Wrappers for BLAS, LAPACK, and Special

insertinterestingnamehere:
Here's a little material on the Cython wrappers. I'm less familiar with the Cython wrappers for scipy.special though, so there may be more to be said about them.

### DIFF
--- a/scipy-1.0/paper.tex
+++ b/scipy-1.0/paper.tex
@@ -314,6 +314,11 @@ The wrappers for BLAS and LAPACK are primarily generated using type information 
 The input and output types of each routine are saved in a data file that is read at build time and used to generate the corresponding Cython wrapper files.
 The wrappers in \texttt{scipy.{\allowbreak}special.{\allowbreak}cython\_special} are also generated from a data file containing type information for the wrapped routines.
 
+Since SciPy can be built with LAPACK 3.4.0 or later, Cython wrappers are only provided for the routines that maintain a consistent interface across all supported LAPACK versions.
+In other words, wrappers are provided for routines that are included in the earliest supported LAPACK version and are not removed or changed in some incompatible way in some subsequent LAPACK release.
+The standard BLAS interface provided by the various existing BLAS libraries is not currently changing, so changes are not generally needed in the wrappers we provide.
+Changes to the Cython wrappers for the functions in \texttt{scipy.{\allowbreak}special} follow corresponding changes to the interface of that submodule.
+
 \subsection*{Numerical optimization}
 \input{scipy-optimize}
 
--- a/scipy-1.0/paper.tex
+++ b/scipy-1.0/paper.tex
@@ -52,6 +52,7 @@
 \author[21]{Scott Sievert}
 \author[22]{Yu Feng}
 \author[23]{Antonio Horta Ribeiro}
+\author[24]{Ian Henriksen}
 
 \affil[1]{Affiliation, department, city, postcode, country}
 \affil[2]{Affiliation, department, city, postcode, country}
@@ -80,6 +81,9 @@
 \affil[21]{Affiliation, department, city, postcode, country}
 \affil[22]{Affiliation, department, city, postcode, country}
 \affil[23]{Affiliation, department, city, postcode, country}
+\affil[24]{University of Texas at Austin,
+           Instite for Computational Engineering and Sciences,
+	   Austin, TX, 78712, USA}
 
 \affil[*]{ralf.gommers@gmail.com}
 
--- a/scipy-1.0/paper.tex
+++ b/scipy-1.0/paper.tex
@@ -300,8 +300,8 @@ from a Cython module using \texttt{scipy.LowLevelCallable.from\_cython}.
 
 \subsection*{Cython bindings for BLAS, LAPACK and special}
 
-When writing algorithms or wrappers in Cython, it is often convenient to call directly into the libraries that SciPy wraps rather than routing the call back through the Python interpreter.
-SciPy currently includes Cython wrappers for many BLAS and LAPACK routines and for many of the special functions provided in the \texttt{scipy.{\allowbreak}special} submodule.
+When writing algorithms or wrappers in Cython\cite{behnel2011cython}, it is often convenient to call directly into the libraries that SciPy wraps rather than routing the call back through the Python interpreter.
+SciPy currently includes Cython wrappers\cite{blas-lapack-wrappers-scipy-2015} for many BLAS and LAPACK\cite{LAPACK} routines and for many of the special functions provided in the \texttt{scipy.{\allowbreak}special} submodule.
 These wrappers are available in the modules \texttt{scipy.{\allowbreak}linalg.{\allowbreak}cython\_blas}, \texttt{scipy.{\allowbreak}linalg.{\allowbreak}cython\_lapack}, and \texttt{scipy.{\allowbreak}special.{\allowbreak}cython\_special} respectively.
 The low-level interfaces for Cython can also be used by third party libraries to get direct access to the functions in these libraries while avoiding the overhead of Python function calls{\textemdash}this can give performance gains of one or two orders of magnitude for many use cases.
 Cython's cimport machinery allows Cython based modules to share C-level objects with other Cython modules without requiring that the third-party extensions link against any additional libraries.
@@ -310,7 +310,7 @@ This lets other extensions avoid the complexity of finding and using the correct
 The low-level Cython wrappers provide an OS and Fortran compiler independent interface for the libraries they wrap.
 
 To help with both correctness and ease of maintenance, most of these low-level Cython wrappers are generated automatically.
-The wrappers for BLAS and LAPACK are primarily generated using type information that is parsed from the BLAS and LAPACK source files using f2py, though a small number of routines use hand-written type signatures instead.
+The wrappers for BLAS and LAPACK are primarily generated using type information that is parsed from the BLAS and LAPACK source files using f2py\cite{peterson2009f2py}, though a small number of routines use hand-written type signatures instead.
 The input and output types of each routine are saved in a data file that is read at build time and used to generate the corresponding Cython wrapper files.
 The wrappers in \texttt{scipy.{\allowbreak}special.{\allowbreak}cython\_special} are also generated from a data file containing type information for the wrapped routines.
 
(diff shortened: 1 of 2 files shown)


insertinterestingnamehere:
If it would help, I can add a few words about how we expose only LAPACK routines that are common across all the versions we support.

tylerjereddy:
Cool, thanks, I hadn't realized the low-level bindings were automatically generated by `f2py` like that.

scipy-1.0/paper.tex
```diff
@@ -258,6 +262,20 @@ \subsubsection*{LowLevelCallable}
 
 \subsection*{Cython bindings for BLAS, LAPACK and special}
 
+When writing algorithms or wrappers in Cython, it is often convenient to call directly into the libraries that SciPy wraps rather than routing the call back through the Python interpreter.
+SciPy currently includes Cython wrappers for many BLAS and LAPACK routines and for many of the special functions provided in the \texttt{scipy.{\allowbreak}special} submodule.
+These wrappers are available in the modules \texttt{scipy.{\allowbreak}linalg.{\allowbreak}cython\_blas}, \texttt{scipy.{\allowbreak}linalg.{\allowbreak}cython\_lapack}, and \texttt{scipy.{\allowbreak}special.{\allowbreak}cython\_special} respectively.
+The low-level interfaces for Cython can also be used by third party libraries to get direct access to the functions in these libraries while avoiding the overhead of Python function calls.
```
rgommers: Because many readers won't know what this means, maybe add something like "- this can give performance gains of one or two orders of magnitude for many use cases".

rgommers:
> If it would help, I can add a few words about how we expose only LAPACK routines that are common across all the versions we support.

That would be useful I think.

scipy-1.0/paper.tex
```diff
@@ -258,6 +262,20 @@ \subsubsection*{LowLevelCallable}
 
 \subsection*{Cython bindings for BLAS, LAPACK and special}
 
+When writing algorithms or wrappers in Cython, it is often convenient to call directly into the libraries that SciPy wraps rather than routing the call back through the Python interpreter.
```
mdhaber: Regarding "call directly into the libraries that SciPy wraps rather than routing the call back through the Python interpreter", does this mean e.g. importing `scipy.special.cython_special` and calling `scipy.special.cython_special.jve(...)` rather than importing `scipy.special` and calling `scipy.special.jve(...)`?

scipy-1.0/paper.tex
```diff
@@ -258,6 +262,20 @@ \subsubsection*{LowLevelCallable}
...
+SciPy currently includes Cython wrappers for many BLAS and LAPACK routines and for many of the special functions provided in the \texttt{scipy.{\allowbreak}special} submodule.
+These wrappers are available in the modules \texttt{scipy.{\allowbreak}linalg.{\allowbreak}cython\_blas}, \texttt{scipy.{\allowbreak}linalg.{\allowbreak}cython\_lapack}, and \texttt{scipy.{\allowbreak}special.{\allowbreak}cython\_special} respectively.
+The low-level interfaces for Cython can also be used by third party libraries to get direct access to the functions in these libraries while avoiding the overhead of Python function calls.
+Cython's cimport machinery allows Cython based modules to share C-level objects with other Cython modules without requiring that the third-party extensions link against any additional libraries.
+Function pointers to wrappers for library routines can also be shared in this way, so that Cython modules using the interface provided by SciPy no longer have to worry about finding and linking against the original library.
```
mdhaber: "the interface provided by SciPy" - does this refer to the same "low-level interfaces for Cython" as in the previous paragraph?

scipy-1.0/paper.tex
```diff
@@ -258,6 +262,20 @@ \subsubsection*{LowLevelCallable}
...
+This lets other extensions avoid the complexity of finding and using the correct libraries on the various possible Fortran compiler ABI and OS combinations.
+The low-level Cython wrappers provide an OS and Fortran compiler independent interface for the libraries they wrap.
+
+To help with both correctness and ease of maintenance, most of these low-level Cython wrappers are generated automatically.
+The wrappers for BLAS and LAPACK are primarily generated using type information that is parsed from the BLAS and LAPACK source files using f2py, though a small number of routines use hand-written type signatures instead.
```
mdhaber: It looks like f2py [should be stylized](https://docs.scipy.org/doc/numpy-1.13.0/f2py/usage.html) as F2PY or \texttt{f2py}.

scipy-1.0/paper.tex
```diff
@@ -258,6 +262,20 @@ \subsubsection*{LowLevelCallable}
...
+The input and output types of each routine are saved in a data file that is read at build time and used to generate the corresponding Cython wrapper files.
+The wrappers in \texttt{scipy.{\allowbreak}special.{\allowbreak}cython\_special} are also generated from a data file containing type information for the wrapped routines.
```
mdhaber: Was the data file for `cython_special` not generated by `f2py`? Was it handwritten? If one of these, it's possible that the information can be slipped into one of the previous sentences rather than being a separate sentence (which could make it more specific/clear and shorter at the same time).

insertinterestingnamehere: Good question. I'm less familiar with the internal layout of `scipy.special` so I was being cautious. As I understand it, the functions there have two interfaces. One is the ufunc interface that can be used from within Python and the other is the low-level Cython interface that can only be used from within Cython. Both sets of wrappers are generated from a single data file (https://github.com/scipy/scipy/blob/master/scipy/special/functions.json) that describes the interface for each function being wrapped. I believe the information there is written by hand though. Perhaps @person142 could clarify more though.

scipy-1.0/paper.tex
insertinterestingnamehere: Thanks for this, I've added text to that effect.

ev-br:
Let's merge this and keep iterating. Thank Ian, all.

ev-br:
Oops, no, there's a merge conflict already.

insertinterestingnamehere:
Thanks for the reviews. I can continue edits elsewhere. Rebased.

scipy-1.0/paper.tex
```diff
@@ -296,6 +300,25 @@ \subsubsection*{LowLevelCallable}
 
 \subsection*{Cython bindings for BLAS, LAPACK and special}
 
+When writing algorithms or wrappers in Cython\cite{behnel2011cython}, it is often convenient to call directly into the libraries that SciPy wraps rather than routing the call back through the Python interpreter.
+SciPy currently includes Cython wrappers\cite{blas-lapack-wrappers-scipy-2015} for many BLAS and LAPACK\cite{LAPACK} routines and for many of the special functions provided in the \texttt{scipy.{\allowbreak}special} submodule.
+These wrappers are available in the modules \texttt{scipy.{\allowbreak}linalg.{\allowbreak}cython\_blas}, \texttt{scipy.{\allowbreak}linalg.{\allowbreak}cython\_lapack}, and \texttt{scipy.{\allowbreak}special.{\allowbreak}cython\_special} respectively.
+The low-level interfaces for Cython can also be used by third party libraries to get direct access to the functions in these libraries while avoiding the overhead of Python function calls{\textemdash}this can give performance gains of one or two orders of magnitude for many use cases.
```
mdhaber: Consider "_These_ low-level interfaces..." to make it clear that "low-level interface" will now be synonymous with "Cython wrapper". 

"libraries" is used to mean very different things in this sentence. Maybe "...third party code" or "extensions" like below.

Rather than the em-dash, I think a new sentence is appropriate.

scipy-1.0/paper.tex
```diff
@@ -296,6 +300,25 @@ \subsubsection*{LowLevelCallable}
...
+Cython's cimport machinery allows Cython based modules to share C-level objects with other Cython modules without requiring that the third-party extensions link against any additional libraries.
+Function pointers to wrappers for library routines can also be shared in this way, so that Cython modules using the interface provided by SciPy no longer have to worry about finding and linking against the original library.
+This lets other extensions avoid the complexity of finding and using the correct libraries on the various possible Fortran compiler ABI and OS combinations.
+The low-level Cython wrappers provide an OS and Fortran compiler independent interface for the libraries they wrap.
```
mdhaber: Perhaps you are making a few significant, but only subtly distinct, points in the preceeding four sentences (307-310), but can't distinguish among them and am somewhat confused. 

I think the central point is that these Cython wrappers allow the use of BLAS, LAPACK, and special functions 1) efficiently (as mentioned before) and 2) conveniently, avoiding complicated linking stuff.

Can this be simplified?

scipy-1.0/paper.tex
```diff
@@ -296,6 +300,25 @@ \subsubsection*{LowLevelCallable}
...
+
+To help with both correctness and ease of maintenance, most of these low-level Cython wrappers are generated automatically.
+The wrappers for BLAS and LAPACK are primarily generated using type information that is parsed from the BLAS and LAPACK source files using F2PY\cite{peterson2009f2py}, though a small number of routines use hand-written type signatures instead.
+The input and output types of each routine are saved in a data file that is read at build time and used to generate the corresponding Cython wrapper files.
+The wrappers in \texttt{scipy.{\allowbreak}special.{\allowbreak}cython\_special} are also generated from a data file containing type information for the wrapped routines.
+
+Since SciPy can be built with LAPACK 3.4.0 or later, Cython wrappers are only provided for the routines that maintain a consistent interface across all supported LAPACK versions.
+In other words, wrappers are provided for routines that are included in the earliest supported LAPACK version and are not removed or changed in some incompatible way in some subsequent LAPACK release.
```
mdhaber: I think this clarification can be removed. The first sentence is sufficient.

Or actually, is this paragraph appropriate for "Key Technical Improvements"? These details about exactly what Cython wrappers are provided and how they are maintained may not be needed here.

scipy-1.0/paper.tex
mdhaber: @person142 If you can answer this, I think it would help to merge this sentence with the one about BLAS and LAPACK.

mdhaber:
@insertinterestingnamehere Just wanted to call your attention to further review comments I posted here after the merge.

insertinterestingnamehere:
@mdhaber Thanks for the additional comments. I'll revise shortly.

scipy-1.0/paper.tex
insertinterestingnamehere: Yah, this one can be simplified. The idea I was expressing is that cimporting these wrappers relies on Cython shuttling a bunch of function pointers around through the Python interpreter instead of just linking against the original library. Strictly speaking, that's not the only way to set up Cython wrappers, but the level of detail there obscures the main points about why the wrappers are useful in the first place.

insertinterestingnamehere:
@mdhaber https://github.com/scipy/scipy-articles/pull/58 should hopefully address your additional comments.

PR #55, opened 2018-07-26
Update the number of public API objects from 2600 to 1500.

rgommers:
See gh-50 for details.


rgommers:
Trivial fix, merging. Thanks @ev-br for pointing out a better method to obtain this number.

PR #56, opened 2018-07-26
Add citation for chain mesh vs. k-d tree performance

tylerjereddy:
The cited manuscript mentions a mesh algorithm outperforming a tree-based code in its abstract, which seems relevant to the discussion here.


tylerjereddy:
@rainwoodman Is this citation ok for chain mesh being faster sometimes?

rainwoodman:
Sorry. It is not a correct citation. That paper is about mesh refinement computing gravity. This context is pair-counting.

I don't think anyone particularly written paper about comparing speed of pair-counters.

Here is a benchmark done by the Corrfunc (chaining mesh) authors: https://gist.github.com/manodeep/cffd9a5d77510e43ccf0

tylerjereddy:
@rainwoodman Noted, thanks

PR #58, opened 2018-07-27
Further Cleanup in Cython Wrapper Section

insertinterestingnamehere:


scipy-1.0/paper.tex
```diff
@@ -298,25 +298,26 @@ \subsubsection*{LowLevelCallable}
...
+When writing algorithms in Cython, it is typically more efficient to call directly into the libraries SciPy wraps rather than indirectly, using SciPy's Python APIs.
+These low-level interfaces for Cython can also be used outside of the SciPy codebase to gain access to the functions in the wrapped libraries while avoiding the overhead of Python function calls.
+This can give performance gains of one or two orders of magnitude for many use cases.
+
+Developers can also use the low-level Cython interfaces without linking against the wrapped libraries\cite{blas-lapack-wrappers-scipy-2015}.
+This lets other extensions avoid the complexity of finding and using the correct libraries.
+Avoiding this complexity is especially important when wrapping libraries written in Fortran.
+Not only can these low-level wrappers can be used without a Fortran compiler, they can also be used without having to handle all the different Fortran compiler ABI's and name mangling schemes.
```
rgommers: grammar thingy here, "can ... can be used"

insertinterestingnamehere: Fixed. Thank you!

scipy-1.0/paper.tex
```diff
@@ -298,25 +298,26 @@ \subsubsection*{LowLevelCallable}
...
-Cython's cimport machinery allows Cython based modules to share C-level objects with other Cython modules without requiring that the third-party extensions link against any additional libraries.
-Function pointers to wrappers for library routines can also be shared in this way, so that Cython modules using the interface provided by SciPy no longer have to worry about finding and linking against the original library.
-This lets other extensions avoid the complexity of finding and using the correct libraries on the various possible Fortran compiler ABI and OS combinations.
-The low-level Cython wrappers provide an OS and Fortran compiler independent interface for the libraries they wrap.
-
-To help with both correctness and ease of maintenance, most of these low-level Cython wrappers are generated automatically.
+\subsection*{Cython bindings for BLAS, LAPACK, and special}
+
+SciPy includes Cython\cite{behnel2011cython} wrappers for many BLAS and LAPACK\cite{LAPACK} routines and for the special functions provided in the \texttt{scipy.{\allowbreak}special} submodule.
```
stefanv: Is Cython explained elsewhere in the paper?  Otherwise, we need an introduction that tells the reader what Cython in, why it is being used (not three paragraphs down, as is currently the case).  Something along the lines of:

Cython is a Python to C compiler, utilized by SciPy to gain speed improvements and to wrap low-level C and C++ libraries.  For example, ... [ckdtrees, special, blas] ...

The sentence later "When writing algorithms in Cython, it is typically more efficient..." can then be removed.

rgommers: @insertinterestingnamehere that suggestion makes sense to me

rgommers:
@insertinterestingnamehere if you're referring to https://github.com/scipy/scipy-articles/pull/54#discussion_r205622080, then that comment is reasonable. I don't think this discussion on LAPACK routines takes too much space though, so only edits for style/clarity are appropriate, no need to try and squeeze the content to reduce word count.

scipy-1.0/paper.tex
jarrodmillman: It can introduce Cython in the intro/history section.  I was already planning to mention it, so it would make sense to explain it as well.

insertinterestingnamehere: That seems like a good idea. Perhaps we can introduce it there and then elaborate a little more on why we use it in an independent paragraph in the language choices section?

insertinterestingnamehere:
@rgommers yes, https://github.com/scipy/scipy-articles/pull/54#discussion_r205622080 is what got me thinking on this. I already removed the redundant sentence mentioned in that comment. If we're not hurting for space, I'm inclined to leave the paragraphs about wrapper generation and ABI evolution where they are because they add meaningful information specifically about this feature.

rgommers:
> If we're not hurting for space, I'm inclined to leave the paragraphs about wrapper generation and ABI evolution where they are because they add meaningful information specifically about this feature.

Sounds good.

scipy-1.0/paper.tex
rgommers: Sounds good. Seems like there aren't any changes needed to this section then.

rgommers:
Merged, thanks @insertinterestingnamehere, all

PR #59, opened 2018-07-27
Address comment of @mdhaber on Impact Now section.

rgommers:


scipy-1.0/paper.tex
```diff
@@ -552,11 +552,11 @@ \section*{Discussion}
...
-From the user side, there were 13,096,468 PyPI downloads
-of SciPy during the year 2017\cite{pypinfo}, which is likely a vast
-underestimate of total downloads given that we don't have access to the
-popular Anaconda package manager download statistics. The recommended 
-source for citing SciPy is simply a reference to the website / library
+From the user side, there were 13,096,468 PyPI downloads of SciPy from the
+Python Packaging Index (PyPI) during the year 2017\cite{pypinfo}, which is
+a lower bound on the number of total downloads by users given that downloading
+from PyPI is only one of several popular methods to install SciPy. The
+recommended source for citing SciPy is simply a reference to the SciPy website
```
ev-br: On a quick glance it looked like a recommendation not to keep citing
the website.

scipy-1.0/paper.tex
```diff
@@ -552,11 +552,11 @@ \section*{Discussion}
 consists of 610 unique contributors of source code, with $>19,000$ commits
 accepted into the code base (GitHub page data).
 
-From the user side, there were 13,096,468 PyPI downloads
-of SciPy during the year 2017\cite{pypinfo}, which is likely a vast
-underestimate of total downloads given that we don't have access to the
-popular Anaconda package manager download statistics. The recommended 
-source for citing SciPy is simply a reference to the website / library
+From the user side, there were 13,096,468 PyPI downloads of SciPy from the
```
tylerjereddy: can remove this mention of PyPI now that it is properly defined / mentioned later in the same sentence

scipy-1.0/paper.tex
tylerjereddy: That wasn't really my original intention with the sentence; perhaps just to indicate that even though we don't have a single clear "proper" journal article or proceedings that should be cited for SciPy, a somewhat informal source like a URL has nonetheless been cited a fair bit.

Not sure if it is a bad thing to discourage citation of the URL moving forward if / when a paper with more inclusive authorship is published; the conservative thing is just to encourage citation of both sources.

rgommers: I see the problem with this phrasing. 

Rewording to make clear that the website is only cited because there was never a paper. 

> the conservative thing is just to encourage citation of both sources.

I don't think so. There should be one recommended citation in places like https://scipy.org/citing.html, which should be this paper not the website.

rgommers:
comments addressed.

rgommers:
This is a minor tweak, merging to keep the ball rolling (slowly this week).

PR #62, opened 2018-08-15
Consistency

jarrodmillman:
This is mostly cosmetic.  I used section/subsection/subsubsection consistently and wrapped all lines.  I added a fixme command and used it.  And I ran a spell checker and confirmed each change.


rgommers:
> I suspect we could add line wrap & spell checkers to the CI process if we really wanted -- perhaps overkill though.

Indeed overkill I think. Let's get this in. Thanks Jarrod, Tyler.

PR #63, opened 2018-09-08
Cite astropy in broader community section draft

tylerjereddy:
I read through the recent community-focused [astropy manuscript](http://iopscience.iop.org/article/10.3847/1538-3881/aabc4f) and thought it might be suitable to at least cite them in a community-related section of our paper.

There are common authors on both projects (and I'm not one, so that removes some bias), and I think there has been some decent exchange of ideas / things that worked well between the two connected communities.

They cite NumPy and SciPy as building blocks in their manuscript, and several conference proceedings where @stefanv and @jarrodmillman were editors, etc., so community connection.

A *specific* example of transfer back to SciPy, if there is one that comes to mind, might make this more compelling if others think it is a good idea--this may not be the right subsection for the content though.

### DIFF
--- a/scipy-1.0/paper.tex
+++ b/scipy-1.0/paper.tex
@@ -592,7 +592,15 @@ functions, parabolic cylinder functions, and spheroidal wave functions.
 
 \subsection*{Community beyond the SciPy library}
 
-\fixme{TODO}
+The scientific Python ecosystem includes many examples
+of domain-specific software libraries building on top
+of SciPy features, and then returning to the base SciPy library
+to suggest and even implement improvements to enable
+more effective science applications. For example, there
+are common contributors to the SciPy and astropy core
+libraries\cite{astropy-2018}, and what works well for 
+one of the code bases, infrastructures, or communities 
+is often transferred in some form to the other.
 
 \subsection*{Maintainers and contributors}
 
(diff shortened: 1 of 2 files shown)


jakevdp:
> A specific example of transfer back to SciPy, if there is one that comes to mind, might make this more compelling if others think it is a good idea

Not precisely an astropy->scipy thing, but the ``binned_statistics`` functionality is something I first developed in an astropy-affiliated package, and then contributed to scipy: #173

tylerjereddy:
Revised to include example form @jakevdp. If this is a subsection the community wants we may want to balance out with an example from one more sub-community as well, but I'll leave that open for feedback.

scipy-1.0/paper.tex
```diff
@@ -592,7 +592,18 @@ \subsection*{Roadmap}
...
+The scientific Python ecosystem includes many examples
+of domain-specific software libraries building on top
+of SciPy features, and then returning to the base SciPy library
+to suggest and even implement improvements to enable
+more effective science applications. For example, there
+are common contributors to the SciPy and astropy core
+libraries\cite{astropy-2018}, and what works well for 
+one of the code bases, infrastructures, or communities 
+is often transferred in some form to the other. At the code
+base level, the \texttt{binned\_statistics} functionality
```
apbard: what do you think about: 

At the code base level, the \texttt{binned\_statistics} functionality _is one of such cross-project contributions: it_ was ...

just to clarify that this is an "example of the example"

tylerjereddy:
@apbard Revised, thanks

tylerjereddy:
Ok, no objections after 12 days so I'll merge it in. Ralf or others can always ask me to remove it when we go through the draft later.

Issue #65, opened 2019-01-11
Revisions Tracking Issue


tylerjereddy:
Based on a full read-through of the manuscript on January 11/ 2019, my detailed revision notes/ check-list is below & of course open to adjustment, etc. Some of the objectives of the revisions at this stage include unifying the paper to a "coherent voice" / style, looking at concision / focus due to its length, and considering the [submission requirements](https://www.nature.com/srep/publish/guidelines) of the journal *Scientific Reports* as well.

- [ ] format our manuscript for submission based on the Journal format requirements:
- > "Scientific Reports publishes original research in one format, Article. In most cases we do not impose strict limits on word counts and page numbers, but we encourage authors to write concisely and suggest authors adhere to the guidelines below."
  - > Articles should be no more than 11 typeset pages in length
  - > the main text (not including Abstract, Methods, References and figure legends) should be no more than 4,500 words.
  - > The manuscript text file should include the following parts, in order: 
    - > a title page with author affiliations and contact information (the corresponding author should be identified with an asterisk). 
    - > The main text of an Article can be organised in different ways and according to the authors' preferences, it may be appropriate to combine sections.
    - > Figure Legends (these are limited to 350 words per figure) 
    - > Tables (maximum size of one page)
   - [x] Footnotes are not used.
  - > authors may choose to incorporate the manuscript text and figures into a single file up to 3 MB in size in either a Microsoft Word, LaTeX, or PDF format - the figures may be inserted within the text at the appropriate positions, or grouped at the end.

**Notes / check-list as I read through the paper**:
- **Introduction**  [my general thought on the intro section is that it is rather anecdotal & would benefit from at least bolstering with some more citations -- i.e., cite conference proceedings instead of just mentioning social events / community & so on; also may consider effectively fusing with Background, which is also anecdotal]
  - [x] add citation / link for the published SciPy-importing LIGO scripts mentioned
  - [x] in the first paragraph we should immediately add some kind of citation for NumPy & explain briefly what it is for the audience of the journal
  - [x] citation for "...backed by millions in funding, and thousands of highly qualified engineers..."
  - [x] citation for "...entire ecosystem of related packages..." &
  - [x] "...a variety of social activities centered around them" (cite conferences?? it is done a bit later, but either do it first here or remove)
- **Background** [anecdotal, but mostly with appropriate citations -- some subsections are noted to need some development and / or cleanup + one or two are abrupt changes in direction from the prose that precedes them]
[...]

PR #69, opened 2019-01-14
Add SI Figures for Sparse Performance

tylerjereddy:
Checklist item from #65.

* add / cite SI Figures for our sparse
performance improvements

However, it doesn't look like the `asv` results substantiate the second performance claim, that CSR to COO matrix format conversion is now several times faster--looks about the same to me for the results I've pulled off Pauli's website. @perimosocordiae any ideas?

As previously noted, I'm also slightly concerned about using rigid screencaps for our figures, but a workflow to generate custom asv figures would be a pain I suppose. Just have to keep in mind that the fonts / proporitions, sizes, axis labels, and so on may not look as professional as one might like so we may get asked to adjust that.

Supporting Info should probably be a separate PDF eventually -- same PDF here for convenience only.


perimosocordiae:
I'm trying to remember which scipy.sparse optimizations are relevant here. I've dug through some old PRs to find:

- scipy/scipy#5946 : asformat {DIA,COO} -> {CSR,CSC}
  - https://pv.github.io/scipy-bench/#sparse.Conversion.time_conversion?p-to_format='csr'&p-to_format='csc'&x-axis-scale=date&p-from_format='coo'&p-from_format='dia'&python=2.7&Cython=0.27.3
 - scipy/scipy#5928 : DIA.diagonal
   - https://pv.github.io/scipy-bench/#sparse.Diagonal.time_diagonal?p-format='dia'&y-axis-scale=log&Cython=0.27.3&python=2.7&numpy=1.8.2
 - scipy/scipy#7081 : CSR/CSC toarray
   - https://pv.github.io/scipy-bench/#sparse.Densify.time_toarray?p-format='csr'&p-format='csc'&x-axis-scale=date&numpy=1.8.2&y-axis-scale=zoom
   - it's a bit noisy, but there's a clear improvement for row-major + CSR and col-major + CSC.

There's also a decent speedup from scipy/scipy#7827 for sparse matrix slicing, but it'll be included in the future 1.3.0 release:
 - https://pv.github.io/scipy-bench/#sparse.NullSlice.time_100_cols?p-format='csr'&p-format='csc'&y-axis-scale=log&Cython=0.29.2

tylerjereddy:
Updated to replace the unconvincing (second) benchmark plot with one of CJ's suggestions, and adjust the text accordingly for the performance claim.

Can we get away with the first of these plots? It is still a demonstration of improvement, but a grumpy reviewer may argue that just around the time of the 1.0 release things started regressing a bit performance-wise.

perimosocordiae:
IIRC, the slight regression was due to correctness fixes unrelated to the original speed improvement.

tylerjereddy:
Ok, I should briefly mention that for clarity I think, thanks.

tylerjereddy:
There's an approving review from a core developer & 6 days without any objections after that. PR has been open 16 days total.

Merging.

This means we have SI section now -- may eventually want to build the LaTeX so it is a separate PDF, or we can just split it off for submission maybe.

PR #70, opened 2019-01-14
Add citation for LIGO SciPy code usage

tylerjereddy:
Checklist item from #65 

* adds citation for the LIGO open [code page that imports the SciPy modules](https://www.gw-openscience.org/s/events/GW150914/GW150914_tutorial.html) described in the body of our manuscript
* since the above page is documented as no longer maintained, also cite the "The LIGO Open Science Center," which has a [more permanent citation target from a conference series](http://iopscience.iop.org/article/10.1088/1742-6596/610/1/012021/meta). Looks like this is what they consider important to cite.

There are also more well-maintained i.e., Jupyter notebooks / Binder setups floating around, but they seem to have disclaimers that they weren't actually used for the manuscript proper.

Anyway, hopefully these two citations should suffice--certainly, we can't expect the reviewers to be Googling around for this stuff in the absence of a citation.


mdhaber:
The sentence is unclear as to whether it claims that SciPy was used in the analysis of the data published in the primary papers, e.g. [1](https://journals.aps.org/prl/pdf/10.1103/PhysRevLett.116.221101), [2](https://journals.aps.org/prl/abstract/10.1103/PhysRevLett.118.221101). The citations we have prove that SciPy is used in analysis scripts geared toward disseminating the LIGO work, but I think we need more to indicate that SciPy is used in the core work.

For instance, I see that [SciPy is a dependency of PyCBC](https://github.com/gwastro/pycbc/blob/master/setup.py), [PyCBC's site](https://pycbc.org/) states that "PyCBC was used in the first direct detection of gravitational waves by LIGO and is used in the ongoing analysis of LIGO and Virgo data", and indeed the paper about [GW150914](https://core.ac.uk/download/pdf/84146534.pdf) cites PyCBC's GitHub page.

tylerjereddy:
Yeah, any help would be appreciated -- I definitely agree with the view that much of the work available for coding / citation is for demonstration rather than the publication proper. I suspect whoever wrote the original prose (Jarrod or Stefan) would like to associate with the primary publication if possible, since that gives us some "scientific weight."

mdhaber:
I have found a few scipy imports in the PyCBC source:
```
# Handle stupid scipy imports (fixed in scipy 0.13)
scipy.factorial = scipy.misc.factorial
```
`from scipy.interpolate import interp1d`
```
hiddenimports = [...
                 'scipy.linalg.cython_blas',
                 'scipy.linalg.cython_lapack',
                 'scipy.special._ufuncs_cxx',
                 ...
```
`from scipy import signal`
(uses kaiserord, get_window, lfilter, zpk2sos, sosfilt, firwin)
`from scipy import interpolate`
(uses interp1d, UnivariateSpline, RectBivariateSpline)
`from scipy.io.wavfile import write as write_wav`
`import scipy.optimize`
(uses fsolve, bisect, brentq, minimize)
`import scipy.stats`
(uses `gaussian_kde`, `multivariate_normal`, `entropy`, `ks_2samp`, `kstest`, `truncnorm.rvs`, `norm.cds`)
`from scipy.integrate import quad`
`from scipy import special`
(uses `logsumexp`, `i0`)

mdhaber:
Based on that, I think we can cite any (and all?) papers that cite PyCBC and also cite the PyCBC source code. Together, they are solid evidence that SciPy has been used in the core gravitational wave analysis.

tylerjereddy:
Updated to cite PyCBC using their preferred method, which is the Zenodo DOI. If we know exactly which algorithms were used I suppose we could also cite the papers that describe those algorithms, which PyCBC also mentions as a component of proper citation.

tylerjereddy:
I used the latest Zenodo DOI rather than i.e., the one that might have been used for the manuscript proper, but I suspect that is ok.

mdhaber:
The LIGO papers should mention which algorithms from PyCBC were used, but I don't think that we have to repeat that information. 
So I think the PyCBC citation is good. Should we also cite one of the LIGO results papers that cites PyCBC (e.g. https://core.ac.uk/download/pdf/84146534.pdf) to establish the link between SciPy and LIGO?

tylerjereddy:
Revised to add that additional citation & resolve merge conflicts.

We don't actually cite the "big" LIGO discovery paper until later in the manuscript for whatever reason.

duncanmmacleod:
all, I can help with a bit more information, but I haven't found the time to sit down and think about it properly, I will try and do that in the next couple of days if that's ok. Just wanted to let you know I am following this.

mdhaber:
Let's merge this for now. @duncanmmacleod, if you can add to this portion, please reply to this when you can and/or create a PR.

PR #76, opened 2019-01-16
Convert footnotes to proper references

tylerjereddy:
Checklist item from #65 -- *Scientific Reports* prohibits footnotes, so use proper references instead.

We're now over double the suggested cap reference count:

> As a guideline references should be limited to 60 (this is not strictly enforced).

But let's deal with that if / when it becomes an issue. This is representing a big body of work / history, after all.

### DIFF
--- a/scipy-1.0/paper.tex
+++ b/scipy-1.0/paper.tex
@@ -220,7 +220,7 @@ concepts such as memory management.
 
 In 1995, Jim Hugunin, a graduate student from MIT, wrote in the first
 message on the Python Matrix Special Interest Group (Matrix-SIG)
-mailing list\footnote{https://mail.python.org/pipermail/matrix-sig/1995-August/000002.html}:
+mailing list\cite{Hugunin-first}:
 \begin{quote}
 There seems to be a fair amount of interest in the Python community
 concerning the addition of numeric operations to Python.  My own desire is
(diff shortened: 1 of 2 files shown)


scipy-1.0/paper.tex
```diff
@@ -828,9 +830,8 @@ \subsection*{API and ABI evolution}
 significant) costs to users, \textit{and} only after giving clear deprecation
 warnings to those users for at least one year.  In general, we encourage
 changes that improve clarity in the API of the library but strongly discourage
-breaking backwards compatibility\footnote{
-    Backwards compatibility means that new versions of SciPy will keep user
-    code written for an older version working.}
+breaking backwards compatibility (Backwards compatibility means that new
```
mdhaber: Parentheses in the middle of a sentence usually don't include a full sentence, right? I think this would go after the sentence.  But then again, perhaps we can assume that the reader is familiar with the notion of backwards compatibility? It is not a technical term, really. And I'm sure there are loads of other places where familiarity with more technical stuff is assumed, so I'd just remove this.

scipy-1.0/references.bib
```diff
@@ -1275,3 +1275,140 @@ @Book{schroeder:2006:VTK
 edition = 4,
 isbn = {978-1-930934-19-1},
 }
+
+@online{Hugunin-first,
```
mdhaber: Thank you for doing this. I won't check all of them at this time; I'd prefer to check through all the references at once as we near the end.

scipy-1.0/paper.tex
tylerjereddy: Agreed, removed.

scipy-1.0/references.bib
tylerjereddy: ok, rebase / force-push after resolving the merge conflicts in references

PR #79, opened 2019-01-16
Benchmark section revisions

tylerjereddy:
Targeting the following checklist items from #65:

- Benchmark suite
  - the "see above" comment re: unit tests may now be below? 
  - shouldn't the "python run.py" command just be "asv run"? maybe it was supposed to be runtests.py or the custom SciPy handler, but using asv directly is probably preferred anyway
  - "the documentation" should probably be "the asv documentation" in last paragraph

### DIFF
--- a/scipy-1.0/paper.tex
+++ b/scipy-1.0/paper.tex
@@ -973,7 +973,7 @@ the only multivariate discrete distribution available in SciPy.
 The airspeed velocity (asv\cite{asvref}) library enables benchmarking Python packages
 over their lifetimes, and the performance of the SciPy code base was monitored
 with asv starting in February of 2015 (PR \#4501). In addition to ensuring that
-unit tests are passing (see above), confirming that performance generally
+unit tests are passing (see below), confirming that performance generally
 remains constant or improves over the commit hash history of the project allows
 us to objectively measure that our code base is improving, to empower
 scientific applications.
@@ -981,8 +981,9 @@ scientific applications.
 Consider the asv benchmark results shown in Figure~\ref{fig:asvbench}, spanning
 roughly nine years of project history. These demonstrate the gradual
 performance improvements in a nearest-neighbor search through
-\texttt{scipy.spatial.cKDTree.query()}, and can be run using the command:
-
+\texttt{scipy.spatial.cKDTree.query()}, and can be run using the
+following command, where \texttt{run.py} is a light wrapper
+script around \texttt{asv}, packaged with SciPy:
 
 \texttt{python run.py run -e -s 800 --bench "\textbackslash
 btime\_query\textbackslash b" "02de46a546..b3ddb2c"}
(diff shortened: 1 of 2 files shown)


scipy-1.0/paper.tex
```diff
@@ -974,16 +974,17 @@ \subsubsection*{Benchmark suite}
 The airspeed velocity (asv\cite{asvref}) library enables benchmarking Python packages
 over their lifetimes, and the performance of the SciPy code base was monitored
 with asv starting in February of 2015 (PR \#4501). In addition to ensuring that
-unit tests are passing (see above), confirming that performance generally
+unit tests are passing (see below), confirming that performance generally
```
mdhaber: Figure \ref{fig:asvbench}

scipy-1.0/paper.tex
```diff
@@ -974,16 +974,17 @@ \subsubsection*{Benchmark suite}
...
 Consider the asv benchmark results shown in Figure~\ref{fig:asvbench}, spanning
 roughly nine years of project history. These demonstrate the gradual
 performance improvements in a nearest-neighbor search through
-\texttt{scipy.spatial.cKDTree.query()}, and can be run using the command:
-
+\texttt{scipy.spatial.cKDTree.query()}, and can be run using the
+following command, where \texttt{run.py} is a light wrapper
+script around \texttt{asv}, packaged with SciPy:
 
 \texttt{python run.py run -e -s 800 --bench "\textbackslash
```
mdhaber: I'd suggest a comma at the end of this line, and _then_ "where \texttt{run.py} is a light wrapper
script around \texttt{asv} that is packaged with SciPy."

scipy-1.0/paper.tex
```diff
@@ -1004,8 +1005,8 @@ \subsubsection*{Benchmark suite}
 Any pull request can be compared against the \texttt{master} branch with the
 command \texttt{asv continuous master new-feature}. This will provide a
 benchmark against the master branch and the branch a new feature is implemented
-in. More features are available in the documentation, including arguments to
-select which benchmarks to run.
+in. More features are available in the \texttt{asv} documentation, including
```
mdhaber: Even though asv has already been cited, I think it would be appropriate to cite the asv documentation itself here.
https://asv.readthedocs.io/en/stable/index.html

tylerjereddy: done

scipy-1.0/paper.tex
tylerjereddy: done

scipy-1.0/paper.tex
tylerjereddy: This I don't get -- the "see below" refers to the Test Suite subsubsection below, not the figure.

mdhaber: My mistake, but this is all the more reason to reference that section than to leave it ambiguous. I think it would help to give the section a label and reference it just as you would a figure.

mdhaber: If you'd prefer I can merge as-is and add this when I take a pass.

PR #80, opened 2019-01-18
Substantiate competition resources in intro

tylerjereddy:
For checklist item in #65:

- citation for "...backed by millions in funding, and thousands of highly qualified engineers..."

@jarrodmillman Provided one of the citations, and the other I dug up for ESRI, which was reporting > 1 billion revenue 10 years ago in the cited article (some overlap with spatial stuff in SciPy, albeit domain focused; they attend SciPy tutorials / PyCon, etc.).

I tried to find good historic revenue numbers for Wolfram Research, but didn't come up with solid sources. Can be a little tricky for privately held companies in the US if they don't disclose during press interviews as for the cases above I guess.


tylerjereddy:
Revised to adjust with "many hundreds" & add the Mathematica citation, as requested. I had to dig up the citation via UC libraries as it was paywalled. The BibTeX style is showing my custom "Note," which it probably shouldn't, but maybe that's another matter / issue to be resolved.

PR #81, opened 2019-01-24
More accurate description of CPython & citations

tylerjereddy:
For checklist item in #65:

> Backgroud: first paragraph -- one thought here: this is a scientific journal, so we could mention that Python is a dynamically typed language & also cite the most popular CPython implementation with the highly-cited Python - C API reference manual produced by Guido

### DIFF
--- a/scipy-1.0/references.bib
+++ b/scipy-1.0/references.bib
@@ -1510,3 +1510,22 @@ year = 2019,
 url = {https://asv.readthedocs.io/en/stable/index.html},
 urldate = {2019-1-22}
 }
+
+% this reference manual has been cited
+% > 300 times, so perhaps suitable here
+% for CPython
+@online{cpython-refman,
+author = {{Guido van Rossum}},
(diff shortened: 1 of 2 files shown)


scipy-1.0/references.bib
```diff
@@ -1510,3 +1510,22 @@ @online{asvdocs
 url = {https://asv.readthedocs.io/en/stable/index.html},
 urldate = {2019-1-22}
 }
+
+% this reference manual has been cited
+% > 300 times, so perhaps suitable here
+% for CPython
+@online{cpython-refman,
+author = {{Guido van Rossum}},
```
tylerjereddy: There's an editor on this whitepaper too, so not sure if "online" is the correct category or not. An argument may also be made that we should just cite the more recent community-generated manuals in the docs.

mdhaber: I'll review citation details all at once at the end.

PR #85, opened 2019-01-24
Move Roadmap into Future Work subsection

tylerjereddy:
For checklist items from #65:

- Future development
  - Condense the PyData sparse content a little maybe & update if appropriate given any recent roadmap / plan adjustments?
  - there is nothing else in this section -- perhaps just refer back to the roadmap then, or combine roadmap and future directions somehow??

I think the easiest solution here is to remove the "Future Development" section--it is redundant with the preceding "Roadmap" section, which already correctly mentions Hameer's sparse package.

Roadmap section may need updates, but on the other hand we would technically be referring to the roadmap from 1.0 instead of *now*, so just a short table citing the real thing may suffice there anyway (separate checklist item).

### DIFF
--- a/scipy-1.0/paper.tex
+++ b/scipy-1.0/paper.tex
@@ -1096,7 +1096,63 @@ broad diversity inclusion statement, and provide instructions for contacting
 the three members of our Code of Conduct Committee or an external
 representative on the NumFOCUS Board of Directors.
 
-\subsection*{Roadmap}
+
+\subsection*{Community beyond the SciPy library}
+
+The scientific Python ecosystem includes many examples
+of domain-specific software libraries building on top
+of SciPy features, and then returning to the base SciPy library
+to suggest and even implement improvements to enable
+more effective science applications. For example, there
+are common contributors to the SciPy and astropy core
+libraries\cite{astropy-2018}, and what works well for 
+one of the code bases, infrastructures, or communities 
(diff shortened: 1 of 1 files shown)


mdhaber:
This is a tough one. The roadmap section makes sense where it is, and the future development section was like "future work" which is usually in a discussion toward the end of a paper. Could the roadmap section be moved toward the end and referred to as "future work"?

stefanv:
On January 24, 2019 18:07:33 Matt Haberland <notifications@github.com> wrote:
> This is a tough one. The roadmap section makes sense where it is, and the 
> future development section was like "future work" which is usually in a 
> discussion toward the end of a paper. Could the roadmap section be moved 
> toward the end and referred to as "future work"?
That works.

tylerjereddy:
Ok, I moved the roadmap content into the Future Work subsection based on feedback above, instead of the original intent of deleting the Future Work section.

Depending on decisions / merge orders, converting the roadmap to a table in a separate PR may generate some merge conflict issues, but we can deal with that if / when it happens.

tylerjereddy:
Ok, I've followed the suggestions of two core developers here and no other comments in the last week so will self-merge in the spirit of keeping the revisions moving forward so we can submit sooner than later.

mdhaber:
Thanks again. Been a very busy week.

PR #86, opened 2019-01-24
Condense list-like roadmap mini-paragraphs into a Table

tylerjereddy:
For checklist item in #65:

> Roadmap: this still reads like an awkward list -- should it be a Table or fully fleshed out prose?

I've gone ahead and condensed the list-like part of this section into a Table. If people don't like this they can close the PR & flesh the prose out a bit more.

Another thing to think about is whether we cite the devdocs version of roadmap, vs 1.2.0 version of roadmap (what we currently do), or the version of the roadmap at the 1.0 release point. And also, how closely the table has to match the roadmap we cite.

In any case, we've already delieverd on some of these roadmap items, which is a good indicator that they were priorities after 1.0 & that our roadmap does indeed contain things that *do* happen.


stefanv:
It may be worth citing both, one to show where we were at 1.0, the other to show our current thinking (which will remain a good place for people to look, in the future, if they want this information).

tylerjereddy:
Revised to cite 1.0 and devdocs roadmaps, though we hardly need more citations at over double the recommended limit.

PR #91, opened 2019-01-28
Start cleaning up cKDTree citations / accuracy

tylerjereddy:
Starting to tackle a tougher checklist item from #65:

> cKDTree [this section needs major revision / clarification / TODO items completed / some material moved to SI?]

There are detailed sub-items related to the above checklist item -- I'll let the reviewer check those, but this does not aim to solve all of them--mostly focused on improving the accuracy and citations in the second paragraph of this section, and removing some stuff that didn't belong.

The first sentence was completely wrong--we didn't add k nearest neighbor searches to cKDTree in 2015. It was there at least two years prior, and the general pure-Python query() algorithm was added in 2008 by @aarchiba. Looks like Anne cited a 1999 paper for the algorithm on the original commit, and we're still using that same rough citation in our docstring for cKDTree, so I've included that here along with the empirical time complexity assessment performed by @jakevdp in 2013 & an updated equivalent I've now performed for SciPy 1.0.0 with the same code / notebook. I think a refined version of that whitepaper (?) was published in a journal / conference proceeding in 2001, which may be worth checking out for the docstring, but maybe out of scope here.

This is at least the second PR to propose addition of Supporting Information, so merge conflicts will invariably pop up too.


scipy-1.0/ckdtree.tex
```diff
@@ -2,22 +2,20 @@
...
+Specifically, in 2015, we enhanced \texttt{cKDTree.query} by reimplementing in C++,
+removing some of the algorithm memory leaks, and allowing 
+release of the GIL so that multiple threads may be used\cite{gh-4374}. In 2013,
+the time complexity of the k nearest neighbor search from \texttt{cKDTree.query}
+was approximately loglinear \cite{knn-jake}, and we've confirmed that this time
+complexity has been preserved in SciPy \texttt{1.0.0} (Figure~\ref{fig:knn-complexity}),
+but with generally improved efficiency (Figure~\ref{fig:asvbench}). The
+discussion in the originally-cited nearest neighbor search
+algorithm publication\cite{kdtree-search-algo}, suggests that there are
+at least some scenarios where $O(log n)$ search time complexity may
```
tylerjereddy: LaTeX formatting for the time complexity could be improved

mdhaber: [This](https://texblog.org/2014/06/24/big-o-and-related-notations-in-latex/) suggests 
`$O(n\log{}n)$ % regular O`
or
`$\mathcal{O}(n\log{}n)$ % Open at top left`

scipy-1.0/ckdtree.tex
```diff
@@ -2,22 +2,20 @@
...
-are all interrogated, but only those neighbors that are requested are retained
-in memory. Thus, the maximum heap size for this algorithm is conceptually determined
-as \texttt{np.arange(max(k))}.
+Specifically, in 2015, we enhanced \texttt{cKDTree.query} by reimplementing in C++,
+removing some of the algorithm memory leaks, and allowing 
+release of the GIL so that multiple threads may be used\cite{gh-4374}. In 2013,
+the time complexity of the k nearest neighbor search from \texttt{cKDTree.query}
+was approximately loglinear \cite{knn-jake}, and we've confirmed that this time
+complexity has been preserved in SciPy \texttt{1.0.0} (Figure~\ref{fig:knn-complexity}),
+but with generally improved efficiency (Figure~\ref{fig:asvbench}). The
```
mdhaber: Maybe reversing the order of these would flow better?
```
In 2013, the time complexity of the k nearest neighbor search from \texttt{cKDTree.query}
was approximately loglinear \cite{knn-jake}. Since then, we've enhanced \texttt{cKDTree.query} by reimplementing in C++, removing memory leaks, and allowing release of the GIL so that multiple threads may be used\cite{gh-4374}, generally improving performance on any given problem (Figure~\ref{fig:asvbench}) while preserving the asymptotic complexity (Figure~\ref{fig:knn-complexity}).
```
I can't see the figures right now, so I hope that's what was intended.
Re: using multiple threads - for this problem, is it possible for multiple threads on a single to core to improve performance, or should we be more specific e.g. "so that multiple cores can be used simultaneously".

scipy-1.0/ckdtree.tex
```diff
@@ -2,22 +2,20 @@
...
+discussion in the originally-cited nearest neighbor search
+algorithm publication\cite{kdtree-search-algo}, suggests that there are
+at least some scenarios where $O(log n)$ search time complexity may
+be achieved, but the influence of dimensionality appears to be consistent
+with our empirical observations of approximately $0(n log n)$ performance.
```
mdhaber: I think this is referring to the part in the reference:
"The seminal work in this area was by Friedman, Bentley, and Finkel [21] who showed that O(n) space and O(log n) query time are achievable for fixed dimensional spaces in the expected case for data distributions of bounded density through the use of kd-trees. There have been numerous variations on this theme. However, all known methods suffer from the fact that as dimension increases, either running time or space increase exponentially with dimension."
But before reading this in the reference, this part of the paper sounded to me like there might be a problem with our implementation. 
Is the intention of citing the paper to give evidence that O(n log n) is reasonable?

scipy-1.0/ckdtree.tex
apbard: +1
```suggestion
at least some scenarios where $\mathcal{O}(\log{n})$ search time complexity may
```

scipy-1.0/ckdtree.tex
tylerjereddy: I removed these specifics based on your proposed refactoring above, since "roughly loglinear" seems to flow nicely enough while preserving some concision.

scipy-1.0/ckdtree.tex
tylerjereddy: Yeah, I've used something similar to your suggestion in the latest set of revisions here now.

I think theoretically releasing the GIL can empower usage of multiple threads on a single physical core that has multiple logical cores for use with hyperthreading, but in practice that may not have been the primary motivation. I haven't changed that specific detail, for now.

scipy-1.0/ckdtree.tex
```diff
@@ -2,22 +2,16 @@
...
-neighbors to account for (i.e., \texttt{[1, 3, 4]}), the intervening distances
-between the nearest neighbor ($k = 1$) and the farthest neighbor requested
-are all interrogated, but only those neighbors that are requested are retained
-in memory. Thus, the maximum heap size for this algorithm is conceptually determined
-as \texttt{np.arange(max(k))}.
+In 2013, the time complexity of the k nearest neighbor search from
+\texttt{cKDTree.query} was approximately loglinear \cite{knn-jake},
+consistent with its formal description \cite{kdtree-search-algo}.
+Since then, we've enhanced \texttt{cKDTree.query} by reimplementing in
+C++, removing memory leaks, and allowing release of the GIL so that
```
apbard:
```suggestion
C++, removing memory leaks, and allowing release of the global interpreter lock (GIL) so that
```

PR #92, opened 2019-01-29
Improvements to cKDTree paragraph 4

tylerjereddy:
Checklist item from #65:

> sentence starting "The cKDTree module implements a dual tree counting algorithm..." is a bit sudden / list-like

Specific improvements include dates / PR links for feature additions to clarify what is actually new, and usage of the actual function / method name for clarity. We could also add `d0afcaf6a` for the initial addition of `count_neighbors`, though I've just added the year for that since it is not a recent improvement. The `cKDTree` docstring is consistent with the primary literature citation originally used, so that should be okay. The use of that algorithm is described as "loose," but good enough I think.

### DIFF
--- a/scipy-1.0/ckdtree.tex
+++ b/scipy-1.0/ckdtree.tex
@@ -34,8 +34,7 @@ operations on the sparse matrix, SciPy allows the dictionary of keys
 sparse matrix to be directly converted to the common CSR, CSC, and COO
 data structures.
 
-The cKDTree module implements a dual tree counting algorithm\cite{Moore2000ar},
-with an improvement to the pair counting algorithm to improve the scaling
-with the number of bins. The cKDTree can now be augmented by weights, with 
-weighted paircount essential in many scientific applications, e.g. computing 
-correlation functions of galaxies\cite{0004-637X-750-1-38}.
+In 2015 the cKDTree dual tree counting algorithm\cite{Moore2000ar}
+was enhanced to support weights\cite{ckdtree-weights}, which are
+essential in many scientific applications, e.g. computing correlation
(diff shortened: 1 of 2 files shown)


stefanv:
Nothing wrong with these changes per se, but I am concerned that such changes add even more detail to an already dense paper.  Perhaps a one or two word addition to the original phrasing would be enough to clarify the "now".

Thoughts, @mdhaber?

mdhaber:
The update clarifies what is new and gives a needed citation. But I @stefanv's point. Perhaps it can be simplified without losing the important information about what's new by removing the first part altogether. Maybe just:

cKDTree now also supports weights in its pair counting routine \cite{ckdtree-weights}, which is essential in 
many scientific applications, e.g. computing correlation functions of 
galaxies \cite{0004-637X-750-1-38}.

apbard:
if we would like to keep the quote for Moore2000ar
maybe something like this?

In 2015, cKDTree dual tree counting algorithm \cite{Moore2000ar} has been enhanced to support weights \cite{ckdtree-weights}, which is essential in
many scientific applications, e.g. computing correlation functions of
galaxies \cite{0004-637X-750-1-38}.

tylerjereddy:
Revised to match suggestion from @apbard, rebased on latest master, resolved merge conflicts, and force pushed.

mdhaber:
@tylerjereddy I have usually resolved merge conflicts myslef and then merged, but I'm a bit confused by the merge conflicts here. What is supposed to be kept?
If you can resolve them I'd be happy to merge.

tylerjereddy:
I rebased and resolved the merge conflicts. It does get messy when parts of the section under modification get deleted, etc., but the diff looks clean again, to me.

PR #93, opened 2019-01-31
Remove extraneous content near end of KDTree section

tylerjereddy:
Meant to address checklist item from #65:

> There's a TODO-style sentence in this section with "Add a Figure to show the scaling, before and after. perhaps give an example or some formula. (cite / mention faster implementions of paircounting algorithms / treecorr, corrfunc)"

This PR proposes to remove that TODO-style content, and we're already providing the correct citation for the main algorithm if the reader wants to explore the literature in greater detail. The docstring for `cKDTree` also contains citations for example applications and example formula, etc. Plus Figure 4 is actually already a cKDtree performance improvement plot.

Alternatively, the content could be moved to the supporting information and filled in there, but unless someone wants to do that very soon, I suggest just removing given length / interest constraints.


mdhaber:
I'll merge in a few days if there are no objections.

PR #94, opened 2019-01-31
Remove common infra change subsection

tylerjereddy:
For checklist item from #65:

> Common Infrastructure
  This subsection is blank -- do we want to keep it?

This PR proposes removing it. Alternatively, we could shift to supporting information, but only if someone is proposing specific content or willing to write it.


rgommers:
Agreed, not too important so I'm fine with leaving it out.

PR #95, opened 2019-01-31
Cite SciPy Developer Guide

tylerjereddy:
For checklist item in #65:

> Maintainers and contributors
  citation / link for SciPy Developer Guide maybe

It is hardly desirable to add more citations, and if that is the decision we can close this and check off the item. Otherwise, this directs the reader to our dev guide for contributing.


mdhaber:
If we want to mention the developer guide, it deserves a citation, so this is helpful. We can consider removing mention of the developer guide once we see how everything fits together.

PR #96, opened 2019-02-01
Improvements to cKDTree paragraph 3

apbard:
I have tried to shrink a bit paragraph 3.
I also removed the reference to `(sets of points organized into a binary space 
partitioning data structure\cite{Bentley:1975:MBS:361002.361007})` ... if we think we need a definition of what a kdtree is, maybe it is better to put it at the beginning of the section.


scipy-1.0/ckdtree.tex
```diff
@@ -18,27 +18,20 @@
...
-the dictionary mapping. The C++ level sparse matrix construction releases the Python
-global interpreter lock for increased performance. Once constructed, the
-dictionary of keys sparse matrix has the amortized constant time complexity 
-($O(1)$) for distance value retrieval that one would expect from a 
-dictionary\cite{Cormen:2001:IA:580470}. For efficiently performing arithmetic
-operations on the sparse matrix, SciPy allows the dictionary of keys
-sparse matrix to be directly converted to the common CSR, CSC, and COO
-data structures.
+In 2015, SciPy added support for generating approximated sparse distance matrices 
+between \texttt{KDTree} objects by zeroing all distances that exceed a user-provided value.
```
mdhaber: I haven't used this, so maybe it's right as written but...
I read "exceed" as "greater than". I would expect values _less than_ a threshold to be set to zero.

scipy-1.0/ckdtree.tex
```diff
@@ -18,27 +18,20 @@
...
-efficient for matrix construction because all zero values are skipped in
-the dictionary mapping. The C++ level sparse matrix construction releases the Python
-global interpreter lock for increased performance. Once constructed, the
-dictionary of keys sparse matrix has the amortized constant time complexity 
-($O(1)$) for distance value retrieval that one would expect from a 
-dictionary\cite{Cormen:2001:IA:580470}. For efficiently performing arithmetic
-operations on the sparse matrix, SciPy allows the dictionary of keys
-sparse matrix to be directly converted to the common CSR, CSC, and COO
-data structures.
+In 2015, SciPy added support for generating approximated sparse distance matrices
```
mdhaber: "approximated" is the past tense verb; I think we want the adjective "approximate"

scipy-1.0/ckdtree.tex
```diff
@@ -18,27 +18,20 @@
...
-global interpreter lock for increased performance. Once constructed, the
-dictionary of keys sparse matrix has the amortized constant time complexity 
-($O(1)$) for distance value retrieval that one would expect from a 
-dictionary\cite{Cormen:2001:IA:580470}. For efficiently performing arithmetic
-operations on the sparse matrix, SciPy allows the dictionary of keys
-sparse matrix to be directly converted to the common CSR, CSC, and COO
-data structures.
+In 2015, SciPy added support for generating approximated sparse distance matrices 
+between \texttt{KDTree} objects by zeroing all distances that exceed a user-provided value. 
+The distance metric used between k-d trees is not constrained to the
```
mdhaber: Does this mean that the routine for generating the approximate sparse matrices is flexible with regard to the distance metric used by the k-d trees? If so, please mention the routine that generates the sparse matrices somewhere in the sentence.  When I first read this, I thought the sentence was just about k-d trees and didn't see the connection to the routine that generates the sparse matrices.

scipy-1.0/ckdtree.tex
```diff
@@ -18,27 +18,20 @@
...
-operations on the sparse matrix, SciPy allows the dictionary of keys
-sparse matrix to be directly converted to the common CSR, CSC, and COO
-data structures.
+In 2015, SciPy added support for generating approximated sparse distance matrices 
+between \texttt{KDTree} objects by zeroing all distances that exceed a user-provided value. 
+The distance metric used between k-d trees is not constrained to the 
+conventional L2 (Euclidean) norm---any Minkowski p-norm value bounded by 1 and 
+infinity is valid. By default, the returned data structure is a dictionary of 
+keys based sparse matrix, which is very efficient for matrix construction. 
+Previously-published results had suggested that this hashing approach to sparse
```
mdhaber: I think we're emphasizing how fast the code is, but "Previously-published results had suggested" prepares me for something to the contrary, like "Previously-published results had suggested this would be fast, but [this is not the case...]"
How about "This hashing approach to sparse matrix assembly can be 7 times faster than constructing with compressed row format (CSR) \cite{10.1007/978-3-540-75755-9_107}"

scipy-1.0/ckdtree.tex
```diff
@@ -18,27 +18,20 @@
...
+between \texttt{KDTree} objects by zeroing all distances that exceed a user-provided value. 
+The distance metric used between k-d trees is not constrained to the 
+conventional L2 (Euclidean) norm---any Minkowski p-norm value bounded by 1 and 
+infinity is valid. By default, the returned data structure is a dictionary of 
+keys based sparse matrix, which is very efficient for matrix construction. 
+Previously-published results had suggested that this hashing approach to sparse 
+matrix assembly is 7 times faster than constructing with compressed row format (CSR)
+\cite{10.1007/978-3-540-75755-9_107} and the C++ level sparse matrix construction 
+releases the Python GIL for increased performance. Also, once the matrix is 
+constructed, distance value retrieval has an amortized constant time complexity
```
mdhaber: How about something like?
"Once the matrix is constructed, distance value retrieval has an amortized constant time complexity \cite{Cormen:2001:IA:580470}, and the DOK structure can be efficiently converted to a CSR, CSC or COO matrix to allow for speedy arithmetic operations."

scipy-1.0/ckdtree.tex
```diff
@@ -18,27 +18,20 @@
...
+infinity is valid. By default, the returned data structure is a dictionary of 
+keys based sparse matrix, which is very efficient for matrix construction. 
+Previously-published results had suggested that this hashing approach to sparse 
+matrix assembly is 7 times faster than constructing with compressed row format (CSR)
+\cite{10.1007/978-3-540-75755-9_107} and the C++ level sparse matrix construction 
+releases the Python GIL for increased performance. Also, once the matrix is 
+constructed, distance value retrieval has an amortized constant time complexity 
+\cite{Cormen:2001:IA:580470}. For efficiently performing arithmetic operations 
+on it, SciPy allows the sparse matrix to be directly converted to the common 
+CSR, CSC and COO data structures.
```
mdhaber: We're inconsistent about the use of acronyms. CSR is defined, dictionary of keys is never assigned an acronym, and CSC/COO are not defined. I'd suggest looking at whether all of these have already been defined in the paper, and if not, defining all of them here.

scipy-1.0/ckdtree.tex
tylerjereddy: I think it is fine as is -- effectively ignoring points beyond a certain distance probably narrows the search space between two kdtrees (two sets of points).

The docstring for `scipy.spatial.cKDTree.sparse_distance_matrix` indicates:

> Computes a distance matrix between two cKDTrees, leaving as zero any distance greater than `max_distance`.

scipy-1.0/ckdtree.tex
tylerjereddy: `scipy.spatial.cKDTree.sparse_distance_matrix` allows any Minkowski p-norm between 1 and infinity.

scipy-1.0/ckdtree.tex
mdhaber: Can we try a numerical example?

Let's say `max_distance` is set as 1.
A distance is calculated to be 0.0001. It does not exceed `max_distance`, so it remains 0.0001.
Another distance is calculated to be 1.1. It exceeds `max_distance`, so this distance is converted to 0.

Do I understand correctly?

scipy-1.0/ckdtree.tex
apbard: right

scipy-1.0/ckdtree.tex
apbard: Yes, that means that is not stored in the matrix and as @tylerjereddy said it speed up things (see [here](https://github.com/scipy/scipy/blob/v1.2.0/scipy/spatial/kdtree.py#L907)). I have changed to `ignoring`, better?

tylerjereddy: I think, informally, dividing the points into the binary space partitioning tree means that when you have a distance cutoff you don't even have to bother calculating some of the distances because of the spatial relationships established by the tree leaves.

So you can just leave those ignored distances as 0 in the sparse matrix -- the algorithm gains efficiency and the data structure loses density.

tylerjereddy: Expanding on that a little, if you divide an initial cube into many small cubes that represent leaves on a binary space partitioning tree, and you have a max distance that is no greater than say the width of two cubes, you don't need to look at points that are registered as distant neighbors on the tree (the cubes are known to be too far apart, just set distance to 0), but you do need to look at neighboring cubes & the points within them.

scipy-1.0/ckdtree.tex
apbard: agreed

scipy-1.0/ckdtree.tex
apbard: see if this way is better

scipy-1.0/ckdtree.tex
apbard: agreed

scipy-1.0/ckdtree.tex
apbard: defined here. then we can remove then if we need to define them before

scipy-1.0/ckdtree.tex
mdhaber: Thanks. @tylerjereddy right, I just didn't know that 0 was just being used to mean "infinity" or "ignore". 
I've done this exact sort of thing but I was making a sparse "similarity" graph, where similarity is an inverse measure of distance.  For that it was a little more natural to ignore ("leave at zero") values smaller than a threshold.

Anyway, this is fine now.

scipy-1.0/ckdtree.tex
```diff
@@ -11,28 +11,20 @@
...
-global interpreter lock for increased performance. Once constructed, the
-dictionary of keys sparse matrix has the amortized constant time complexity 
-($O(1)$) for distance value retrieval that one would expect from a 
-dictionary\cite{Cormen:2001:IA:580470}. For efficiently performing arithmetic
-operations on the sparse matrix, SciPy allows the dictionary of keys
-sparse matrix to be directly converted to the common CSR, CSC, and COO
-data structures.
+In 2015, SciPy added the \texttt{sparse\_distance\_matrix} routine for generating 
+approximate sparse distance matrices between \texttt{KDTree} objects by ignoring 
+all distances that exceed a user-provided value. Also, this routine it is not
```
tylerjereddy:
```suggestion
all distances that exceed a user-provided value. Also, this routine is not 
```

tylerjereddy:
Just one more small fix suggested from my end & I think we can merge this helpful revision.

scipy-1.0/ckdtree.tex
apbard: right

tylerjereddy:
In it goes, thanks @apbard

PR #99, opened 2019-02-14
Coverage Analysis / Plot Improvements

tylerjereddy:
For checklist item from #65:

> Figure 5: bizarre dips and lack of x-axis label / our inability to modify the figure easily for reviewers seem problematic
> [ ]  we could try to regenerate this programmatically using coverage code over the commit history of the project, but that may be a bit of an undertaking to get right

I reconstituted line coverage analysis between 2013-ish and the 2017 v1.0 release point for Python & compiled languages. The gory details are available in the docker based approach [here](https://github.com/tylerjereddy/scipy-cov-track).

Going back farther than `v0.12.1` (2012-ish) was proving really painful (well, the whole endeavor was painful, really), so unless we need that I probably won't do it.

Of course, there are various complications with presumed additions of more vendored compiled-language packages over time, the gradual introduction of new features into `runtests.py` means that I had to checkout more modern versions of that file for really old SciPy compiled coverage, and some old versions even required manual symlinking of system `.so` files for historical reasons. Forward-support for modern NumPy / Cython was also compromised as I went back in history, of course. And there's the nose to pytest transition...

Here is the new plot (and old one below it) in case reviewers don't feel like looking at the CI compiled version:

cov

coverage-chart


mdhaber:
This is great. Thank you for taking the time to do this.
The style of the plot is good, and the caption addressed the question I hoped it would (coverage dip in compiled code). 
I wonder if you have the data for the total number of lines of code, too? If so I might experiment with a different way of presenting the data.
But I'll merge this in a week if there's nothing to the contrary.

tylerjereddy:
I didn't record the total lines, but we could probably get that out if we want. I suppose it may be nice to identify the total counts for compiled code for assurance about the decrease there.

I'll see what I can do--it does take quite a while to run this stuff, although should be smoother now that the dependency work is done.

tylerjereddy:
The PR now contains a new text file with the total line data extracted by the respective coverage parsers.

Here's a plot that includes the total line data, but it is too confusing in its current form to use in the manuscript. I'm not sure it contributes that much--let's not forget the impact that i.e., switching to pytest and uses its plugin + parametrized testing might have.

For compiled code I suppose we could try to point any major additions that didn't really warrant comprehensive testing because the upstream package is trusted, etc.

(blue dash = Python total lines, etc.)

cov-tot

mdhaber:
Thanks! Here is something that looks like what I was thinking of doing:
https://plot.ly/~Dana.Short/741.png
I'm thinking maybe a red area representing lines of Python could be stacked on a blue area representing lines of compiled code so that the total area is the number of total lines. Then the bottom portion of each area could be shaded darker representing the proportion of lines covered by tests.
You would only be able to read the number of lines directly; proportion of code covered by test would only be visual. But I think it would look interesting.

perimosocordiae:
@mdhaber Something like this, perhaps? https://docs.google.com/spreadsheets/d/1ubHyIaEY6edDa_2rQKw_cELohRPpWhiU12x-Lg5lp78/edit?usp=sharing

python coverage

compiled coverage

I find that splitting the Python from the compiled code makes the plots easier to interpret.

tylerjereddy:
@perimosocordiae cool

mdhaber:
Well, I agree that you can't read the % coverage for Python code so well when they are all together.
chart
But perhaps that just drives home the fact that it's high.

I think that the plot's information density needs to be higher - under the constraint that it remains easy to interpret - than the original line plot. The separate area plots are better, but I don't think we have space for both. So I would suggest the combined plot with a detailed, well-written caption. I'll let others chime in as to which they prefer, though.

tylerjereddy:
Looking at both, my vote is that we probably can *just* get away with the single combined plot as long as we are quite careful in the caption etc. Let's see what a few others think.

I'm more than happy for pushes directly to this branch to replace my original plot / caption. The code to generate the old / original plot is preserved in the docker repo anyway.

ilayn:
You can get a decent TeX plot with matching fonts via (by changing `\usepackage{pgfplots}` to `\usepackage{pgfplotstable}` ; latter imports the former already)

```TeX
\begin{figure}[H]
\centering
\begin{tikzpicture}[]
\pgfplotstableread[column/ver/.style=string type]{
ver totpyline covpyline uncovpyline totcompline covcompline uncovcompline
v0.12.1 33844 25721 8123 273634  142290 131344
v0.13.3 36638 28944 7694 338926  172852 166074
v0.14.1 37301 30587 6714 277643  147151 130492
v0.15.1 38339 30288 8051 304898  158547 146351
v0.16.1 40094 32075 8019 322398  167647 154751
v0.17.1 42566 34478 8088 340903  170452 170451
v0.18.1 44711 36216 8495 353417  173174 180243
v0.19.1 43823 36373 7450 436900  200974 235926
v1.0.0 106878 92984 13894 462574 208158 254416
}\covtable
    \begin{axis}[
    width=16cm,
    height=7cm,
    enlargelimits=false,
    ymin=0,
    ymax=6e5,
    stack plots=y,%   
    area style,
    xtick=data,
    ytick={2e5,4e5,6e5},
    xticklabels from table={\covtable}{ver},
    scaled y ticks=false,
    y tick label style={/pgf/number format/.cd, fixed, 1000 sep={}},
    legend entries={Compiled (covered),Compiled (uncovered),Python (covered),Python (uncovered)},
    legend style={draw=none,fill=none, cells={align=left},/tikz/every even column/.append style={column sep=3mm}},
    legend columns=-1,
    legend image code/.code={\fill[#1] (0cm,-0.1cm) rectangle (0.2cm,0.1cm);},
    ]
  \addplot+[fill=blue!50] table[x expr=\coordindex, y=covcompline] {\covtable} \closedcycle;
  \addplot+[fill=blue!10] table[x expr=\coordindex, y=uncovcompline] {\covtable} \closedcycle;
  \addplot+[fill=red!70] table[x expr=\coordindex, y=covpyline] {\covtable} \closedcycle;
  \addplot+[fill=red!30] table[x expr=\coordindex, y=uncovpyline] {\covtable} \closedcycle;
\end{axis}
\end{tikzpicture}
\caption{\% coverage of the SciPy code base by unit tests over a five year
time span predeceding the 1.0 release point. The reconstitution of historical
build and test environments was performed using a docker-based approach
\cite{scipy-cov}. Analysis of lines of source code covered is performed 
automatically in our continuous integration suite using the \texttt{pytest-cov}
and \texttt{gcov} libraries. Some decreases in compiled code coverage may reflect
inclusion of additional vendored code.}
\label{fig:coverage}
\end{figure}
```

It should give something like 

image

ilayn:
I didn't polish to the production but I am planning to polish all plots after we are done with the content.

mdhaber:
Does this data conflict with Figure 3 (The breakdown of programming languages...)? 
It shows Python making up ~50% of SciPy code base.

tylerjereddy:
There are all kinds of things that could be going on. You'd have to check the algorithm used by [`linguist`](https://github.com/github/linguist), which among other things: 

> excludes all files that it determines to be binary data, vendored code, generated code, documentation, or are defined as data (e.g. SQL) or prose (e.g. Markdown) languages, whilst taking into account any overrides.

Perhaps excluding vendored code is a pretty big deal in that breakdown vs. the breakdowns we extract from the test coverage plugins. We probably could do a side-by-side comparison of some of the categories to see if i.e., vendored code gets excluded in a major way sometimes. Not sure if that is in scope for us to do though.

`linguist` also conflates Python & Cython, or anything else of the `*.py*` variety, I think. There were some longer discussions about this a while ago. There was no interest in separating Python & Cython when I tried opening a PR to linguist a few months ago, back when I did the original analysis, but I did try to patch in the behavior when I ran `linguist` on SciPy. That probably doesn't account for the magnitude of differences observed, but the hacky patch may not help matters.

I'm having trouble installing `linguist` locally on my mac now because of all the ruby / gem stuff. Feel free to re-run it & inspect the breakdown of files for SciPy 1.0.

mdhaber:
@tylerjereddy OK, I can try to run it.
I'm just trying to make sure there isn't a mistake. 
If I get to that, I'll also try remaking Figure 3 as a pie chart. 
We should then decide whether we really need it, as figures eat up a lot of space in this single-column format, and there is some overlap between the information in Figure 3 and the one in this PR.

tylerjereddy:
We have access to the source code for each approach, but may be time consuming to dissect. We will probably have to check if there's a vendored code exclusion as advertised.

I thought there was already a decision in favor of figure 3--could just shift the one here to the SI before merging if space is an issue.

tylerjereddy:
I think there should be consensus before refactoring Figure 3 to a pie chart.

stefanv:
With compiled code, do you mean auto-generated from Cython?  I'm not sure how much insight can be gained from looking at coverage of the latter.  If we look at coverage of the .pyx files themselves, that's different.  But that could all just be bundled together.

tylerjereddy:
- `linguist` (Fig. 3) runs directly on the source, not the compiled version of code; it is on the top bar of all GitHub repos as well

- For "compiled" in this PR, I mean whatever `gcov` & `gcovr` are scraping in the docker containers / repo I cite above for the analysis; if that isn't enough information to know for sure what is / isn't included, then we'd have to modify the Dockerfile to output the list of the "compiled" files that are considered, and look at that.

Ultimately, both linguist & `gcovr` can produce lists of the files parsed, so we can compare the two if we want to dissect this.

mdhaber:
@tylerjereddy Sorry, I really didn't mean to cause a stir! If there was already consensus on keeping Figure 3 despite some overlap with this figure, please ignore the suggestion.  Re: pie chart version; I will certainly submit a PR rather than merging directly.

mdhaber:
Oiy. I have tried installing linguist using both pip and easy_install on three computers. Fails every time.

tylerjereddy:
I've updated the [scipy-cov-track repo](https://github.com/tylerjereddy/scipy-cov-track) to contain the code adjustments and text file results for parsing line data in SciPy 1.0.0 with:

- linguist (now used in the Docker container as well...)
- the Python coverage plugin used by pytest
- gcov & gcovr, for compiled language source files

Those are the three sources of coverage and / or line count-related information for which there is some contention. The results files are stored at path `container_output` in that repo, and obviously can be (re)-generated portably using Docker commands explained in the README there.

If there are specific questions / concerns about the data we have so far, probably a good idea to take a close look at those files & the approaches used by different libraries. For example, `linguist` will effectively report the bytes of source code vs. lines for %, if that matters.

I can see vendored code in several of the list files mentioned above--if we want to exclude that, then proactive suggestions might include:

- the appropriate mechanism for exclusion that is portable before / after the nose -> pytest transition; even better to try it on the Dockerfile first
- the approach for the different accounting libraries

mdhaber:
Changes are good. If we decide to revisit code analysis we can do that later. Thanks @tylerjereddy

PR #100, opened 2019-02-15
Simplify polynomial interpolation content

tylerjereddy:
Aiming to address some checklist items from #65 relating to polynomial interpolators:

> - "UnivariateSpline and splrep/splev combo" -> combination
> - there's also a fair bit of background for this subsection before we get to the new stuff, but perhaps it is sufficiently convoluted that at least some background is justified
> - I suspect we could make some effort to clarify old vs. new here & to migrate some of the material that is more "user guide / reference" like elsewhere


scipy-1.0/poly.tex
```diff
@@ -28,23 +16,12 @@
...
-%
-Using \code{UnivariateSpline} in new code is not recommended. Given data points,
-the helper function \code{make\_interp\_spline} constructs a \code{BSpline}
-instance, representing a spline which interpolates the data. Interpolated values
-at intermediate points can be computed either via its \code{\_\_call\_\_} method,
-or by calling \code{splev(xnew, BSpline\_object)} --- the \code{splev} routine
-accepts both \code{tck} tuples and \code{Bspline} instances.
-
-Fitting \emph{smoothing} splines to data can currently only be done with FITPACK. Building alternative algorithms is certainly possible, but is still to be implemented.
+for e.g., constructing Bezier curves). \code{BSpline} represents spline
```
tylerjereddy: @ev-br Could you give some dates and / or PR links for these new things added in last 3 years or so?

scipy-1.0/paper.tex
```diff
@@ -1263,6 +1262,48 @@ \section*{Supporting Information}
 \label{fig:sparse-conv}
 \end{figure}
 
+\section{Polynomial Interpolators: Additional Details}
```
tylerjereddy: This section may be a little awkward with various bits and pieces pasted in from the main text now, but maybe not such a big deal now that it has been moved to supporting info. Assuming Evgeni is "ok" with this reduction in length for main text section.

scipy-1.0/poly.tex
ev-br: PPoly & BPoly went in in https://github.com/scipy/scipy/pull/5827  for scipy 0.13
BSpline went in in scipy/scipy#3174 for scipy 0.19
NdPPoly went in for scipy/scipy#5827 for scipy 0.18
(Now that I think of it, it is worth mentioning here, if briefly)
CubicSpline was implemented in scipy/scipy#5653, new in scipy 0.18

ev-br:
> Should probably sign off on any adjustments

I'm fine with these adjustments, thanks for the heavy lifting here and elsewhere in the text! 
Srsly though, being the lead author, you do not need my signoff :-).

scipy-1.0/poly.tex
tylerjereddy: I added in the first two & the last one. I'm confused about the first one though -- it is the same as the third, but for a different version of SciPy.

ev-br: That's a typo, sorry. Edited in the correct PR, which is https://github.com/scipy/scipy/pull/2885

tylerjereddy: Revised to point to Pauli's PR, thanks

scipy-1.0/references.bib
```diff
@@ -1679,3 +1679,27 @@ @online{ckdtree-weights
...
+urldate = {2019-2-15}
+}
+
+@online{scipy-gh3174,
+author = {{Evgeni Burovski}},
+title = {{add b-splines}},
+year = 2013,
+url = {https://github.com/scipy/scipy/pull/3174},
+urldate = {2019-2-15}
+}
```
ev-br: I might be missing something, or might have missed some previous discussions (please point me to one if that's the case): why adding these as references, here and elsewhere? FWIW I don't think it's wanted or needed. A PR is not a single-person affair and is not worthy of an academic citation.

tylerjereddy: I think there was a discussion in support of citing them, or maybe it was just a review comment from @mdhaber or @ilayn? We've also got mailing list messages and all sorts of barely-academic things cited. Some were set as footnotes, but footnotes are barred for the journal we will target first.

I don't really mind if they are removed, but then someone should open an issue, and get a reasonable number of people to make a decision. I'm not sure if the journal will allow those citations anyway.

mdhaber: So the issue is whether to cite GitHub PRs/issues or not?
I don't remember the history but I might have said something in favor of them before.
If they are primary sources, I think the citations should at least be collected so we know where we are getting our facts from. 
Citing websites is often discouraged for good reason, but I think this is if the website is a potentially unreliable secondary (or _n_-ary) source. For what we are writing, websites are often the primary source. Citing primary sources is something that certainly makes sense in many disciplines.
So I would suggest collecting/keeping the citations and removing them if the journal asks us to.

PR #101, opened 2019-02-15
Shift optimize background content for concision of main text

tylerjereddy:
For checklist item from #65:

> **Numerical optimization**
> * [ ]  I suspect we should remove the background on what scipy.optimize does with the list of 7 problems -- we should find a way to make this relatively clear & abstract that information to earlier in the paper when the 16 SciPy submodules are described

This PR proposes to fix the issue by moving the longer background content to the SI & citing that SI content from the subpackages description earlier in the manuscript.

cc @antonior92 

There are more simplifications to the optimize section suggested in the checklist for the purpose of concision, but I'll start with this one for now.

### DIFF
--- a/scipy-1.0/paper.tex
+++ b/scipy-1.0/paper.tex
@@ -1263,6 +1263,18 @@ Some additional roadmap items are summarized in Table~\ref{table:roadmap}.
 \label{fig:sparse-conv}
 \end{figure}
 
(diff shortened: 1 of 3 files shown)


mdhaber:
I agree that the paper is about 1.0 and we need to cut things down. 
I supported writing a lot more about optimize than we needed, intending to trim later. 
Moving to supporting information is great.

PR #102, opened 2019-02-15
Table 1 (optimization) concision / recent changes adjustments

tylerjereddy:
For checklist items from #65:

> - **Nonlinear optimization**
>    - **Local**
> - [ ] Table 1 is nice; the sentence citing it should mention that it includes some of the more recent additions since we're still in the umbrella of key / recent technical improvements
> - [ ] that said, Table 1 also includes additions from as far back as SciPy 0.6 -- which is a longer time span than "last 3 years" for the improvements section as a whole...
>    -  [ ] what about i.e., a dashed vertical line / highlight of some sort dividing "last 3 years" and "before that" in the Table?
>  - [ ] The Table caption is very detailed, but I'm a little concerned about its size -- could we cut it down a bit and move some of the heavy details to i.e., supporting information?

The approach drafted in this PR cuts the Table down to just the recent improvements & abstracts the rest to another Table in the SI.

This provided some opportunity to cut down the caption and table itself in main body of text. There may be more opportunities for simplification, but let's open this for review before I over-optimize the optimize section.

### DIFF
--- a/scipy-1.0/scipy-optimize.tex
+++ b/scipy-1.0/scipy-optimize.tex
@@ -33,7 +33,7 @@ Compared to the previously implemented simplex method, the new interior-point me
 The \texttt{minimize} function provides methods for finding local minima of
 nonlinear optimization problems. It unifies several methods with a common 
 interface in order to facilitate interchanging between solvers with different
-characteristics. Some recent improvements are summarized in Table~\ref{tab:minimize}.
+characteristics. Some recent improvements are summarized in Table~\ref{tab:minimize-si}.
 %% Currently, this introduces a ton of whitespace around a very simple expression (min f(x)). The reader can easily find the standard of meaning of `nonlinear programming', so I believe it should be removed. If we decide that it is appropriate to include, then we should add the entire problem statement - including constraints - and we should also include the entire linear programming problem statement in the previous section.
 % summarizes all the methods available for solving minimization problems of the type:
 %\begin{equation}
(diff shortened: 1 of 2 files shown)


tylerjereddy:
Splitting the table has drawbacks too, of course, making it harder to compare new vs. old, but it was perhaps getting unwieldy.

scipy-1.0/scipy-optimize.tex
```diff
@@ -73,59 +74,41 @@ \subsubsection*{Nonlinear Optimization}
...
-    factorization} & & & &  & & & & \cmark & \cmark &  & \cmark &  & \cmark \\
-    \hline
-    Local convergence& & & & L & S &  L & S & S$^*$ & S$^*$ & Q & S$^*$ & Q & S$^*$  \\
+    factorization} &  & \cmark &  & \cmark \\
     \hline
-    Global convergence & & &  &   & \cmark & \cmark & \cmark & \cmark & \cmark & \cmark & \cmark & \cmark & \cmark  \\
+    Local convergence& Q & S$^*$ & Q & S$^*$  \\
     \hline
-    \makecell{Line-search (LS) or\\ trust-region (TR)} & Neither  & LS &  TR & LS & LS & LS & LS & LS & LS & TR & TR & TR & TR \\
+    \makecell{Line-search (LS) or\\ trust-region (TR)} & TR & TR & TR & TR \\
```
tylerjereddy: All the same here, so could remove this row too--I'll wait for feedback on that though.

mdhaber:
I think the whole table should appear in supporting information. 
But maybe it should only appear there. I think it was very important work to collect all this together and put it in such a nice form, but I don't think there is space for it all - and cutting it in half really doesn't solve the problem given the length of caption that is necessary..

mdhaber:
If we decide to cut down this section and move the table to the end, perhaps something like:
The `minimize` function unifies several methods of finding local minima of nonlinear optimization problems in a common interface. Four new methods for unconstrained optimization were added to `minimize` in recent versions of SciPy: `dogleg`, `trust-ncg`, `trust-exact`, and `trust-krylov`. All are trust-region methods that build a local model of the objective function based on first- and second-derivative information, approximate the best point within a local "trust region", and iterate until a local minimum of the original objective function is reached, but each has unique characteristics that make it appropriate for certain types of problems. For instance, `trust-exact` achieves fast convergence by solving the trust-region subproblem almost exactly, but it requires the second-derivative Hessian matrix to be stored and factored every iteration, which may preclude the solution of large problems ($\geq 1000$ variables). On the other hand, `trust-ncg` and `trust-krylov` are well-suited to large-scale optimization problems because they do not need to store and factor the Hessian explicitly, instead using second-derivative information in a faster, approximate way. A detailed comparison of the characteristics of all `minimize` methods is presented in Table~\ref{tab:minimize}.

tylerjereddy:
I think "we" is basically going to mean you & I for the most part--feel free to push right into this PR if you want, or do you want me to maybe just move the whole table for now & then you follow-up with the prose clean-up after?

I thought about moving the whole table to the SI, but decided against it. I'm happy to do it though--our more realistic objective (I think!) is to have a citation target vs. perfect manuscript & even if we had more people reviewing perfecting prose can be an endless rabbit hole (just like coding!).

mdhaber:
Ok, yeah, it would be great if you'd move the table, then I'll change the text in a separate PR

tylerjereddy:
Ok, refactored to relocate the table to SI rather than trying to split the Table.

scipy-1.0/scipy-optimize.tex
```diff
@@ -33,7 +33,8 @@ \subsubsection*{Nonlinear Optimization}
 The \texttt{minimize} function provides methods for finding local minima of
 nonlinear optimization problems. It unifies several methods with a common 
 interface in order to facilitate interchanging between solvers with different
-characteristics. Some recent improvements are summarized in Table~\ref{tab:minimize-si}.
+characteristics. Characteristics of the different methods are summarized in 
+Table~\ref{tab:minimize-si}.
```
tylerjereddy: maybe ", including some recent improvements"

mdhaber:
Ok. I'll incorporate that idea when I change the text in the section.

PR #104, opened 2019-02-20
Change roadmap summary of ndimage

stefanv:
The roadmap description for `ndimage` recently changed: https://scipy.github.io/devdocs/roadmap-detailed.html

Updated the paper summary accordingly.


tylerjereddy:
I think the Table is meant to reflect the state of the roadmap at 1.0 release point rather than at 1.3.0dev.

Otherwise, the table caption needs to be changed & other entries probably need updating if others agree to use more recent roadmap changes.

mdhaber:
I agree that the table was intended to be the 1.0 roadmap. Could indicate in the caption where the current roadmap can be found.
@rgommers would you make the call on this?

tylerjereddy:
This is a pretty minor adjustment / refinement in priority so I'll just merge it.

I think @mdhaber & I agree that we shouldn't refactor the whole table for 1.3.0dev though.

tylerjereddy:
Thanks Stefan

rgommers:
+1 current state LGTM

PR #111, opened 2019-02-24
Language choices

mdhaber:
I got a bit carried away here. The section was well-written as it was but I thought that the disconnect between the language preference order and the actual composition of SciPy would confuse readers. Many changes revolved around addressing this. 

~~I am also suggesting the use of a pie chart, rather than a bar chart, to depict the breakdown of SciPy. This version of the plot was prepared hastily but I thought it would help to see something before making a decision.~~

(Let's pretend I never posted the pie chart - image removed)


ilayn:
Ooh, we will be hunted down by the Tufte forces; "the only worse design than a pie chart is several of them." -edward tufte 😃

tylerjereddy:
I'm not in favor of switching to a pie chart, so someone should tie break on that

scipy-1.0/paper.tex
```diff
@@ -735,52 +735,52 @@ \subsection*{Submodule organization}
...
-    \caption{The breakdown of programming languages used in the
-             SciPy library determined using the linguist library.
-    	 Small ($<0.5 \%$) amounts of TeX, Matlab, Shell,
-    	 and Makefile are excluded for clarity and mostly
-    	 provide supporting roles in tests, building, and
-    	 documentation.}
-    \label{fig:linguist}
+\centering
+\includegraphics[width=0.5\textwidth]{static/scipy_languages}
+\caption{Composition of SciPy: lines of code in each programming language as determined by the \texttt{linguist} package.
```
tylerjereddy: It is not lines of code, it is bytes; the linguist repo points to this: https://developer.github.com/v3/repos/#list-languages

This may also be a source of confusion in the related PR about coverage vs. the data here.

apbard:
I am slightly in favour of the pie chart

scipy-1.0/paper.tex
mdhaber: Oops, I got the two mixed up. I'll fix that. The idea is to address the apparent discrepancy and try to explain it.

mdhaber:
I didn't know pie charts were so hated. How about a single horizontal bar broken into segments rather than wrapped into a pie? That would take less space, too.

ilayn:
I think we don't even need to make a figure for this. 4 items are quite OK to enumerate in text. The graph adds no additional information.

mdhaber:
I'd be fine with that.

tylerjereddy:
I brought this up in June 2018 in #36 when adding the plot:

> the plot is probably too big and even a table or mention in the text might suffice I suppose

It was merged with two core devs reviewing, but fair enough--go ahead and remove.

ilayn:
Indeed, I think this kind of back and forth will be happening more, once we get closer to the finished product.

rgommers:
I'm fine with with no figure. Otherwise a shrunken bar chart seems appropriate. No pie charts anywhere please:)

tylerjereddy:
Maybe just move the Figure to the Supporting Information and move on.

I don't think the changes to the text are warranted, but it probably reads fine either way, so maybe just merge after moving the Fig.

scipy-1.0/paper.tex
```diff
@@ -735,52 +735,52 @@ \subsection*{Submodule organization}
...
+Small ($<0.5 \%$) amounts of TeX, Matlab, Shell,
+and Makefile (not visible) provide supporting roles in tests,
+building, and documentation.}
+\label{fig:linguist}
 \end{figure}
 
-For implementing new functionality, we have a clear order of language
-preference.  First Python, if performance is not an issue. If it is, then in
```
mdhaber: This are not complete sentences.

scipy-1.0/paper.tex
```diff
@@ -735,52 +735,52 @@ \subsection*{Submodule organization}
...
+\centering
+\includegraphics[width=0.5\textwidth]{static/scipy_languages}
+\caption{Composition of SciPy: lines of code in each programming language as determined by the \texttt{linguist} package.
+Small ($<0.5 \%$) amounts of TeX, Matlab, Shell,
+and Makefile (not visible) provide supporting roles in tests,
+building, and documentation.}
+\label{fig:linguist}
 \end{figure}
 
-For implementing new functionality, we have a clear order of language
```
mdhaber: This paragraph, which is about languages preferences, sat between paragraphs about SciPy's actual composition. The preference order and actual composition are quite different, so I think the material about preference should be moved somewhere either before or after the material about actual composition to avoid mixing these distinct thoughts.

scipy-1.0/paper.tex
```diff
@@ -735,52 +735,52 @@ \subsection*{Submodule organization}
...
-preference.  First Python, if performance is not an issue. If it is, then in
-order of preference: Cython, C, C++, Fortran. The main motivation for this is
-maintainability: Cython has the highest abstraction level and most Python
-developers will understand it. C is also widely known, and easier to deal with
-for the core development team than C++ and especially Fortran.
-
-While it is not surprising that Python is heavily used in SciPy, the usage
```
mdhaber: Perhaps this is meant to address the difference between language preference order and actual composition, but I did not see that in previous reads, so I think it needs to be more explicit.

scipy-1.0/paper.tex
```diff
@@ -801,24 +801,31 @@ \subsection*{Language choices}
 
 Rounding out the top three languages in SciPy is C, which is also extremely
 well-established over several decades\cite{Kernighan:1988:CPL:576122} of
-scientific computing. Alongside its objected-oriented relative C++, C can be
-leveraged in Python after being exposed using the Cython language. Cython has
```
mdhaber: This bit about Cython was in the middle of the paragraph about C. Since C is so much more prevalent than Cython, I think it deserves its own paragraph.

mdhaber:
I've reverted to the old figure rather than removing it entirely. That way this PR can be either accepted or rejected based on contributions to the text of the section. How to handle figure removal can be discussed in a separate PR.

scipy-1.0/paper.tex
tylerjereddy: A lot of C is vendored, while a lot of the Cython is developer contributed in recent years, so I think the argument about relative importance can go either way.

tylerjereddy:
I like fixing the incomplete sentences, but I think the rest of the restructure is a stylistic debate & this PR starts with:

> I got a bit carried away here

> The section was well-written as it was

which I mostly agree with, but we can just merge I think. I'll let someone else do that though--I remain somewhat resistant.

scipy-1.0/paper.tex
tylerjereddy: I don't think they're distinct thoughts, but rather tightly interwoven--we prefer certain languages & simultaneously get huge amounts of code in other languages because of wrapping, and having an intermingled paragraph configuration where C & Cython are discussed as a unit & so on could actually convey this point.

mdhaber:
@rgommers @ilayn could you comment on the writing changes? The figure is no longer part of this PR.

ilayn:
I would suggest putting a clean table as the following (without the journal format)

```latex
\documentclass{article}
\usepackage{booktabs,siunitx}

\begin{document}

\begin{table}%
\centering
\begin{tabular}{p{3cm}S[table-format = <0.1]}
\toprule
Language & {Percent}\\
\midrule
Python  & 49.5 \\
Fortran & 25.6 \\
C       & 19.5 \\
Cython  & 3.0  \\
C++     & 2.3  \\[\defaultaddspace]
{TeX, Matlab, Shell, and Makefile} & <0.5 \\
\bottomrule
\end{tabular}
\caption{Language Composition of SciPy codebase: lines of code in each
programming language as determined by the \texttt{linguist} package.
The last row denotes the tools used in supporting roles in tests,
building, and documentation.}
\label{tab:linguist}}
\end{table}

\end{document}
```

This gives 

image


The percentages don't add up to 100% though.

scipy-1.0/paper.tex
```diff
@@ -764,23 +772,15 @@ \subsection*{Language choices}
...
-maintainability: Cython has the highest abstraction level and most Python
-developers will understand it. C is also widely known, and easier to deal with
-for the core development team than C++ and especially Fortran.
-
-While it is not surprising that Python is heavily used in SciPy, the usage
-distribution of the other languages warrants some discussion. Fortran is an
+Fortran is an
 extremely well-established scientific programming language, both for historical
 reasons and because of its continued excellent
 performance\cite{Koelbel:1993:HPF:562354}. We wrap the FFTPACK Fortran library
```
ilayn: Do we need to establish the excellence of these well-known packages? I would just go about it with 

```
Fortran, despite it is age, is still a high-performance scientific programming language with continued contemporary usage. Thus, we wrap the following excellent field-tested Fortran libraries in order to provide the Python convenience while still benefiting from their respective performance: THISPACK \cite{...}, THATPACK \cite{...} ...
```

We can still give some praising comments but these are kind-of known assets out there.

scipy-1.0/paper.tex
```diff
@@ -801,24 +801,31 @@ \subsection*{Language choices}
...
-SciPy library results from a combination of the usage of powerful performance
-enhancing languages that interact well with Python (i.e., Cython), and the
+Cython has been described as a creole language that mixes the best parts of Python and
+lower-level C / C++ paradigms\cite{behnel2011cython}. We often use Cython
+as a glue between well-established low-level scientific computing libraries written
+in C / C++ and the Python interface offered by SciPy. We also use Cython to enable performance
+enhancements in Python code, especially for cases where heavily used inner
+loops benefit from a compiled code with static typing. 
+
+For implementing new functionality, Python is the still the language of choice. If Python
```
ilayn: My suggestion is something along the lines of

```
Adding new functionalities, while highly desired, requires also some future planning
about the maintenance and integration with the rest of the codebase. Hence the higher
the level of language the better it is for other maintainers. Thus, our first language of 
choice is Python if the performance is not a concern. The next in line for increased 
performance is Cython which still offers some resemblance to Python but still offers 
quite increased performance and then lower level languages such as C, C++, or Fortran...
```

scipy-1.0/paper.tex
rgommers: makes sense to me

rgommers:
> The figure is no longer part of this PR.

It seems like the figure png itself is still in the PR? As is the commented out figure related latex? I'd just throw that out. That and the table suggested by @ilayn can be a separate thing then (I thought I saw the horizontal bar figure somewhere, but don't see an open issue/PR anymore.

The rest seems reasonable, mostly moving some things around. I just read this section in the pdf version, and the order of topics seems fine. Agree with @ilayn that the part discussion how C and Fortran are excellent is probably not necessary.

@mdhaber I suggest to address the remaining open comments and then ping me, happy to do a final review/merge.

rgommers:
I'll try to find some time this week to read through the whole paper; it seems like we're getting really close to finishing up.

mdhaber:
@rgommers @ilayn @tylerjereddy Implemented suggestions from @ilayn seconded by @rgommers. Removed the distasteful png. @ilayn's table will be in a separate PR.

scipy-1.0/paper.tex
mdhaber: @ilayn I wasn't sure I understood the purpose of the change, so I had trouble adapting it. I condensed it to the following:
"When implementing new functionality, maintainability is of primary importance. Consequently,
Python is the still the language of choice. If Python performance is an issue, then we prefer 
the use of Cython followed by C, C++, or Fortran (in that order). "
But I don't see either as clear improvement over what we have. Can we consider this in a separate PR if necessary?

mdhaber:
@ilayn Did you agree with this for the most part? If so, would you merge?

ilayn:
Looks good to me, thanks Matt. Only detail I don't know whether you want to keep the commit history or not so based on your decision you can merge it or squash/merge it.

Issue #97, opened 2019-02-12
Review of text on stats module


chrisb83:
I reviewed the text on `stats`:

I agree with @tylerjereddy (see #65 
) that the text in paper.tex could be shortened. From my point of view, the following are the main points if we speak about distributions in `stats`:

- more than 100 distributions available (~80 continuous, 12 discrete and 10 multivariate)
- consistent framework that implements methods to sample rvs, evaluate the pdf / cdf and to fit parameters for every distribution
- generally relies on specific implementations for each distributions, otherwise defaults to generic methods (such as deriving the cdf from the pdf via numerical integration)

Questions / ideas to shorten the current text:

- comment on random_state: is this needed? I assume most users expect that they can provide a seed.
- multinomial: same as random_state, probably expected that this exists. note: "only multivariate" contradicts statement that there are 12 multivariate
- rv_histogram: do we want to sell this as an important feature? advantage / use case probably not clear from the text

When mentioning other features in subpackages.tex, the scope is quite narrow:

> statistical tests, including Pearson's correlation, Spearman's rank-order correlations, Kendall's tau, chi-squared test and its generalization as the Cressie-Read power divergence, contingency table tests including Fisher's exact test and Mood's median test, and many more; and assorted transformations and statisticsof data

- we could mention more generic test classes such as testing correlation, equality of the mean / median / variance, whether a sample follows a certain distribution etc and give a few well-known examples. (the ones given now sucha s Mood and Cressie-Read are not very well-known, just my opinion)
- what about KDE?
- statistical distances?

Happy to get your view.

tylerjereddy:
You may not get too much feedback -- the subset of devs working on the paper actively is pretty small, but I trust your judgement far more than mine on `stats` content, and I probably ended up drafting some of the stats material, unfortunately!

@mdhaber has been good about reviewing PRs carefully, so I think if you clean up the stats stuff a little he'd be happy to review.

mdhaber:
I would be happy to review. I'm also starting to dig into my own edits now that we've gone through a fair portion of Tyler's list. I haven't read the most recent version of the stats section, but when I do I can let you know what I think about it. If you want to submit a PR before then, though, that also works.

chrisb83:
> If you want to submit a PR before then, though, that also works.

I would be happy to get your review of the current version to see if you agree with my comments above before I create a PR.

tylerjereddy:
[...]

PR #113, opened 2019-03-01
Key technical improvements

mdhaber:


scipy-1.0/paper.tex
```diff
@@ -877,30 +873,28 @@ \subsubsection*{Sparse matrices}
...
+
 \subsection*{Unified bindings to compiled code}
 
 \subsubsection*{LowLevelCallable}
 
 As of SciPy version 0.19, it is possible for users to wrap low-level functions
-in a \texttt{scipy.LowLevelCallable()} object that reduces the overhead for
-calling compiled C functions directly from Python.  Supported low-level
-functions include \texttt{PyCapsule} objects, ctypes function pointers, and
-cffi function pointers. The low level function signature must be consistent
```
mdhaber: It's unclear to me whether/why the details about the function signature are important to include in the paper. The bit about "low level function signature must be consistent with the expectations of the routine it is passed to" seems redundant with "as long as the function call signature matches the specifications". I don't think we need to specify that "The C code may be generated using numba or Cython" as there is no reason for the reader to suspect that the source of the C code matters. The `scipy.ndimage.generic\_filter` example doesn't strengthen the text because the reader is probably unfamiliar with `scipy.ndimage.generic\_filter`.
I'd be happy to help with revisions, though, if this part is really important.

scipy-1.0/paper.tex
```diff
@@ -877,30 +873,28 @@ \subsubsection*{Sparse matrices}
...
+Supported low-level functions include \texttt{PyCapsule} 
+objects, \texttt{ctypes} function pointers, and \texttt{cffi} function pointers. 
+Furthermore, it is even possible to generate a low-level callback function 
+automatically from a Cython module using \texttt{scipy.LowLevelCallable.from\_cython}.
+
+\subsubsection*{Cython bindings for BLAS, LAPACK, and special}
 
 SciPy has provided special functions and leveraged BLAS and
 LAPACK\cite{LAPACK} routines for many years. SciPy now additionally
-includes Cython\cite{behnel2011cython} wrappers for:
```
mdhaber: The colon is not necessary here according to rule 2 of:
http://www.kentlaw.edu/academics/lrw/grinker/LwtaColons.htm

scipy-1.0/paper.tex
```diff
@@ -877,30 +873,28 @@ \subsubsection*{Sparse matrices}
...
-many BLAS and LAPACK routines (added in 2015), and the special functions
```
mdhaber: Comma would be needed if we are separating three or more items with commas, but we're not.

scipy-1.0/paper.tex
```diff
@@ -919,7 +913,7 @@ \subsection*{Cython bindings for BLAS, LAPACK, and special}
 Avoiding this complexity is especially important when wrapping libraries
 written in Fortran.  Not only can these low-level wrappers be used without a
 Fortran compiler, they can also be used without having to handle all the
-different Fortran compiler ABI's and name mangling schemes.
+different Fortran compiler ABIs and name mangling schemes.
```
mdhaber: Some would argue against apostrophe here (not possessive).

scipy-1.0/paper.tex
```diff
@@ -950,10 +944,10 @@ \subsection*{Statistical distributions}
 distributions and a framework to implement additional distributions. Given an
 expression for a probability distribution function (pdf), the framework will
```
mdhaber: Is any of this new? Consider removal except for the new stuff at the end.

scipy-1.0/paper.tex
```diff
@@ -975,73 +968,54 @@ \subsection*{Polynomial interpolators}
...
-benchmark against the master branch and the branch a new feature is implemented
-in. More features are available in the \texttt{asv} documentation\cite{asvdocs}, including 
-arguments to select which benchmarks to run.
-
 \subsubsection*{Test suite}
+\label{sec:testsuite}
 
 The SciPy test suite is orchestrated by a continuous integration matrix that
 includes posix and Windows (32/64-bit) platforms managed by Travis CI and
-appveyor, respectively. Our tests cover Python versions 2.7, 3.4, 3.5, 3.6, and
```
mdhaber: What is new?

scipy-1.0/paper.tex
```diff
@@ -975,73 +968,54 @@ \subsection*{Polynomial interpolators}
...
+roughly nine years of project history.
+
+\begin{figure}[H]
+\centering
+\includegraphics[width=\textwidth]{static/asv_time_query_ckdtree}
+\caption{Airspeed velocity benchmarks for \texttt{scipy.spatial.cKDTree.query}. 
+The results are based on
+Python 2.7 performance on the master branch of the project using NumPy 1.8.2
+and Cython versions 0.27.3, 0.21.1, and 0.18 (for improved backward
+compatibility). Only the $L^2$ (Euclidean) norm is shown here, and to improve
```
mdhaber: This still needs revision; I assume it will done when this plot is formatted to be included in the report.
L^2 norm is not shown here. Elapsed time of something (what, exactly?) is shown.
What data is in the tree? What are the arguments to `query` and what do they mean?

scipy-1.0/scipy-optimize.tex
```diff
@@ -22,65 +22,8 @@ \subsubsection*{Linear Optimization}
...
...rained optimization were added to \texttt{minimize} in recent versions of SciPy: \texttt{dogleg}, \texttt{trust-ncg}, \texttt{trust-exact}, and \texttt{trust-krylov}. All are trust-region methods that build a local model of the objective function based on first- and second-derivative information, approximate the best point within a local ``trust region'', and iterate until a local minimum of the original objective function is reached, but each has unique characteristics that make it appropriate for certain types of problems. For instance, \texttt{trust-exact} achieves fast convergence by solving the trust-region subproblem almost exactly, but it requires the second-derivative Hessian matrix to be stored and factored every iteration, which may preclude the solution of large problems ($\geq 1000$ variables). On the other hand, \texttt{trust-ncg} and \texttt{trust-krylov} are well-suited to large-scale optimization problems because they do not need to store and factor the Hessian explicitly, instead using second-derivative information in a faster, approximate way. A detailed comparison of the characteristics of all \texttt{minimize} methods is presented in Table~\ref{tab:minimize-si}.
```
mdhaber: As discussed in #102

scipy-1.0/scipy-optimize.tex
```diff
@@ -22,65 +22,8 @@ \subsubsection*{Linear Optimization}
...
 
 
 \paragraph{Global Minimization}
-As \texttt{minimize} only seeks a local minimum, some problems require the use of a global optimization routine to find the minimum of the objective function over the set of all decision variable values, especially if the objective function possesses many local minima that other solvers could get stuck in. % e.g. if there is concern that `minimize` could get stuck in an unacceptable local minimum?
-The \texttt{brute} function performs a grid search over the entire domain, and the best grid point is ``polished'' by a local minimizer. However, this approach becomes prohibitively expensive as the problem dimensionality increases.
-% As far as I understand how brute works, the run time does not depend on the number of extrema, as it only polishes the best one
-% , or if the objective function has many extrema. 
-To address this, the
-\texttt{basinhopping} \cite{Wales1997} function uses random perturbation of the decision variables in conjunction with local minimization. The acceptance of the perturbed coordinates is dictated by the Metropolis test typically used in Monte Carlo simulations.
```
mdhaber: `brute` and `basinhopping` are not new.

scipy-1.0/paper.tex
```diff
@@ -848,22 +848,18 @@ \section*{Key technical improvements}
...
-\input{ckdtree}
-
 \subsubsection*{Sparse matrices}
 
 \texttt{scipy.sparse} offers seven sparse matrix data structures,
 also known as sparse formats. The most important ones are the row- 
 and column-compressed formats (CSR and CSC, respectively). 
 These offer fast major-axis indexing and fast matrix-vector multiplication,
-and are used heavily in scikit-learn.
+and are used heavily throughout SciPy and dependent packages.
```
mdhaber: Why only mention scikit-learn? 
Can we cite a fact about a different package without citation?

scipy-1.0/paper.tex
tylerjereddy: I disagree -- the L2 norm is definitely shown in this plot. Is the number 2 highlighted in the left side panel. There are different distance metrics possible for cKDTree and L2 is the one chosen for these results and displayed on the plot.

For the arguments to query, may refer the reader to something more detailed in the SI or simply the asv benchmark code itself.

scipy-1.0/paper.tex
tylerjereddy: If you're not sure, then maybe the issue already opened -- #97 regarding the stats section -- is the place to discuss this first. I think @chrisb83 has opened the issue.

scipy-1.0/paper.tex
tylerjereddy: See #97 for this discussion -- maybe do that there

scipy-1.0/paper.tex
tylerjereddy: All of it -- CI is relatively new

tylerjereddy: See the timeline figure you made

scipy-1.0/paper.tex
tylerjereddy: `numba` and `Cython` are major open source projects that provide useful ways to interface with Python for preformance reasons, so the interop seems relevant to me.

Some cut down is probably fine as you say.

scipy-1.0/paper.tex
tylerjereddy: The general statement you made is probably fine -- I'll just mark this as resolved with your adjustment

scipy-1.0/scipy-optimize.tex
tylerjereddy: great, marking as resolved

scipy-1.0/paper.tex
tylerjereddy: ping @chrisb83

chrisb83: I think the framework to handle distributions is worth mentioning, but let's discuss in 97

scipy-1.0/paper.tex
mdhaber: That's why I asked. The timeline shows Travis CI starting in 2013. The section is about
"key technical improvements made in the last three years". Since we started Travis CI more than three years before the release of SciPy 1.0 in late 2017, I thought it was worth checking about the others.

scipy-1.0/paper.tex
mdhaber: Ah, I see that the L^2 norm was an option set in the tests. But I meant that the phrase "Only the $L^2$ (Euclidean) norm is shown here..." suggests that the lines on the plot represent the L^2 norm of something. This could confuse a reader unfamiliar with either k-d trees or `asv` benchmarking.

mdhaber: I will leave this to another PR.

scipy-1.0/paper.tex
mdhaber: How about 
"As of SciPy version 0.19, it is possible for users to wrap low-level functions
in a \texttt{scipy.LowLevelCallable} object that reduces the overhead of
calling compiled C functions, such as those generated using \texttt{numba} 
or Cython, directly from Python."
?

mdhaber:
@tylerjereddy In retrospect, I should have split this into multiple PRs. Let me know what changes to revert so we can discuss them separately.

- [ ] I think we can resolve the LowLevelCallble section changes here. I suggested a change that would bring back the explicit mention of code generated by `numba` and Cython.
- [ ] I didn't make any sweeping changes to the stats section; we're leaving that to #97. I hope the language adjustments are OK. I can explain the reason for each change in detail if it would help.
- [ ] I agree that the changes to the Benchmark suite section are hard to follow because I re-ordered this and the test suite section. If you are willing to consider them here, the main changes are 1) the removal of the complete command for generating Figure~\ref{fig:asvbench} and 2) the crash course on `asv` use. The rationale is that the other sections aren't so detailed, so these seemed out of place. Just let me know if I should revert, though.
- [ ] GitHub is not doing a great job at identifying the the changes to the Test Suite section; they're pretty minor. The biggest is changing 
"There are more than $13000$ unit tests in the test suite, which is written for usage with the pytest runner, and with 87 \% and 45 \% line coverages for Python and compiled code, respectively at the SciPy 1.0 release point (Figure~\ref{fig:coverage})" 
to 
"There are more than $13,000$
unit tests in the test suite, which uses the \texttt{pytest} framework, and coverage of SciPy's approximately $204,000$ lines of code
has held steady at 76.5\% for the last six months (Figure~\ref{fig:coverage})"
This new sentence was based on an old version of the text, though. I was reviewing from a hard copy that didn't include some of the most recent changes. Perhaps it should be:
"There are more than $13,000$ unit tests in the test suite, which uses the \texttt{pytest} framework, with
 87 \% and 45 \% line coverages for Python and compiled code, respectively, at the SciPy 1.0 release point (Figure~\ref{fig:coverage})"
Let me know if I need to revert.
- [ ] I revised the `minimize` section considerably, but we had discussed this text in #102 
- [ ] I removed mention of `brute` and `basinhopping` from Global Minimization. GitHub shows a lot of green and red because I also removed a lot of commented-out lines.

So perhaps we can resolve most changes here, but I'll move the changes to "Test Suite" and "Benchmarking" to a separate PR?

mdhaber:
Split into separate PRs

PR #114, opened 2019-03-03
Language Breakdown Figure

mdhaber:
PR for changing the language breakdown figure. 

Some have suggested removing the figure entirely. Please also suggest how the text should be changed.

In the meantime, here's an alternative figure that takes relatively little space.

image


rgommers:
Agreed that this figure is an improvement over the old one. I'm also fine with no figure - leaving the decision up to people with a stronger opinion.

tylerjereddy:
I would have preferred to just move the other one to SI, but providing constant resistance won't move the paper forward so I'll just merge.

It looks like the "C" logo is actually "C++" unless there's something wrong with my eyes! Maybe fix that up in a follow-up if you agree with me

ilayn:
Oops seems like I've commented in the wrong place but https://github.com/scipy/scipy-articles/pull/111#issuecomment-469584143

I also don't have a strong opinion. In case of confusion I try to go with the cleanest..

PR #116, opened 2019-03-05
Test suite

mdhaber:
Mostly stylization (capitalization, teletype) fixes.
Also added detailed caption to figure proposed in #99.

Please consider:

- "with the `pytest` runner" -> "with the `pytest` framework" (or do we mean "with `pytest-runner`"?)
- moving the line "Some of the historical components of the codebase would still benefit from increased test coverage." to the discussion, when it is written

### DIFF
--- a/scipy-1.0/paper.tex
+++ b/scipy-1.0/paper.tex
@@ -992,7 +992,7 @@ arguments to select which benchmarks to run.
 \subsubsection*{Test suite}
 
 The SciPy test suite is orchestrated by a continuous integration matrix that
-includes posix and Windows (32/64-bit) platforms managed by Travis CI and
+includes POSIX and Windows (32/64-bit) platforms managed by Travis CI and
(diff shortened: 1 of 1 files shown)


mdhaber:
@tylerjreddy code base and codebase are equally valid, definitely. But we need to stay consistent. There was some mixing, so I've made everything I've seen codebase.

mdhaber:
@tylerjereddy 
Re: discussion at the end or by section, I'm not sure. In either case, though, would it be OK to move "Some of the historical components of the codebase would still benefit from increased test coverage." out of this paragraph?

Did "the pytest runner" mean "`pytest-runner`"?
If not, can I change this to "`pytest` framework"? (Readers like me might not know what "runner" means.)

tylerjereddy:
Ok, I merged it.

I meant the code that executes the tests, not the framework for writing tests, so the original was probably more accurate. This is especially true since we probably did / do have some left-over old school testing framework code in our test suite that is reasonably compatible with the pytest runner machinery, but that hasn't been modernized to leverage all that the pytest framework has to offer.

PR #117, opened 2019-03-05
Benchmark Suite

mdhaber:
Moved benchmark suite after test suite because benchmark suite section refers to unit test suite section. Easier for reader to remember the previous paragraph than look ahead. Also a few stylization fixes.

Please consider:

1. removing 1) the command to generate the figure (or moving it into the caption) and 2) and the material after the figure about how to use `asv`. To me it seemed like a short tutorial on how to use `asv` with SciPy, which seemed out of place. If the intent is to advertise the ease of generating such results - which is a great new feature - a description of the capabilities of the `run` script would work
2. removing the last part of "allows us to objectively measure that our code base is improving~~, to empower scientific applications.~~"
3. rewording "generally remains constant or improves over the commit hash history of the project" -> "improves over time."
4. rewording "the performance of the SciPy codebase _was_ monitored with `asv` _starting in_" -> "the performance of the SciPy codebase _has been_ monitored with `asv` _since_"
5. moving the thought "in addition to ensuring that unit tests are passing" to the beginning of the paragraph, which would be right after the unit tests paragraph - a nice segue. 

Perhaps:
In addition to ensuring that unit tests are passing, it is important to confirm that the
the performance of the SciPy codebase improves over time. Since February 2015, the 
performance of SciPy has been monitored with Airspeed Velocity (`asv` \cite{asvref}). 
SciPy's `run.py` script conveniently wraps `asv` features such that benchmark results 
over time can be generated with a single console command. For example, 
Figure~\ref{fig:asvbench} illustrates the improvement of `scipy.spatial.cKDTree.query` 
over roughly nine years of project history.

### DIFF
--- a/scipy-1.0/paper.tex
+++ b/scipy-1.0/paper.tex
@@ -1026,16 +1026,16 @@ inclusion of additional vendored code.}
 
 \subsubsection*{Benchmark suite}
 
-The airspeed velocity (asv\cite{asvref}) library enables benchmarking Python packages
+The airspeed velocity (\texttt{asv}\cite{asvref}) library enables benchmarking Python packages
 over their lifetimes, and the performance of the SciPy code base was monitored
-with asv starting in February of 2015 (PR \#4501). In addition to ensuring that
+with \texttt{asv} starting in February of 2015 (PR \#4501). In addition to ensuring that
 unit tests are passing (see Test suite section below), confirming 
 that performance generally
 remains constant or improves over the commit hash history of the project allows
 us to objectively measure that our code base is improving, to empower
 scientific applications.
 
-Consider the asv benchmark results shown in Figure~\ref{fig:asvbench}, spanning
+Consider the (\texttt{asv} benchmark results shown in Figure~\ref{fig:asvbench}, spanning
 roughly nine years of project history. These demonstrate the gradual
 performance improvements in a nearest-neighbor search through
 \texttt{scipy.spatial.cKDTree.query()}, and can be run using:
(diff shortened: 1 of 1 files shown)


scipy-1.0/paper.tex
```diff
@@ -1065,6 +1024,47 @@ \subsubsection*{Test suite}
 
 % NOTE: is there a citation for PyPy?
 
+\subsubsection*{Benchmark suite}
+
+The airspeed velocity (\texttt{asv}\cite{asvref}) library enables benchmarking Python packages
+over their lifetimes, and the performance of the SciPy code base was monitored
+with \texttt{asv} starting in February of 2015 (PR \#4501). In addition to ensuring that
+unit tests are passing (see Test suite section below), confirming
```
tylerjereddy: The test suite section isn't below anymore so the parenthetical note can be removed. We could probably just not mention the unit test suite in the benchmark section rather than moving the benchmark section, but I don't really care if you prefer this ordering.

mdhaber: I did forget to remove the parenthetical note. Fixed.

mdhaber:
@pv I see you are an author of `asv`. When referring to it by its full name in the article, would you prefer that it be capitalized or not (i.e. "airspeed velocity" or "Airspeed Velocity"? Both stylizations appear on the [PyPI page](https://pypi.org/project/asv/).

tylerjereddy:
I don't think we need to spend core dev cycles on the preferred capitalization of "airspeed velocity." Both are used on its GitHub page.

If others think removing the example `asv` command is useful, then in my view it should be relocated to the SI in the spirit of reproducibility.

mdhaber:
Moving `asv` command to SI is fine. But unfortunately, even with the command I wasn't able to reproduce the results. Besides having a working SciPy development setup and having `asv` installed, are there other requirements? (Other than that, it looks like the command is supposed to set up virtual environments for me?) 

I was going to try to produce a figure from the data. I agree with your initial inclination in #65 that the screenshot needs to be replaced. If you can send the raw data, I can draft a figure.

tylerjereddy:
I don't think we have the raw data anymore, but we do have a sample config file used to generate it at path `/scipy-1.0/supporting_info/asv_bench/cKDTree/asv.conf.json`. If the `asv` run fails there's usually a traceback & you can go from there until it works.

It was quite a bit of work with multiple runs using different dependency versions to accumulate the json results files before opening the html performance page locally. I think I spent a few days getting that just right.

I did ask about how much raw data / detail we should store in the SI. The discussion in #15, almost a year ago now, seems to have 1 vote in favor of the code needed to reproduce, but not the raw data.

I think we should just try with the screenshot now, despite our mutual reservations there, and if there's a request for revision then we can suffer through the regeneration of data if needed.

There's also the website Pauli has for asv performance, but even that data isn't quite right for this purpose since there are a lot of breaks where the dependency versions need to change for builds to succeed and so on.

mdhaber:
In the meantime, digitizing the data from the image is an option. I did a crude job, and the dates aren't properly lined up, but it could be done more carefully if the data is too difficult to reproduce.

Figure

I would have done version numbers instead of dates but I couldn't really read the version numbers in the original, whereas dates could of the first and last commit commit are easy to find and the rest can be interpolated.

mdhaber:
The existing changes are just moving the section and correcting capitalization, stylization, etc. I'll move the rest to separate PRs.

PR #120, opened 2019-03-06
Language Breakdown Figure -> Table?

mdhaber:
Exhibit A (#114 )
image

Exhibit B (this PR, as suggested by @ilayn in #111)
image

Votes?


mdhaber:
@ilayn tests complain about siunitx. Thoughts?

ilayn:
The package is not installed somehow even though it should be [a part of TeXLive ](https://ctan.org/pkg/siunitx?lang=en)

jarrodmillman:
I prefer the table.

tylerjereddy:
Oy, are we still debating this thing!

ilayn:
Parking the debate aside for a moment, it looks like the .circleci file requires 

```bash
- command: sudo apt install texlive-scientific
``` 
after line 8 `checkout`. I used `siunitx` for aligning the second column at the decimal point but if it turns out to slow things down I can manually do that.  It's not really a blocking issue. But if we have a lot of units to be entered this package is the de facto standard.

About the figure, table thingy; my argument is that the programmer audience would definitely know these languages thus don't need the logo and the non-programmer audience won't recognize the logos. Hence either way it seems like they don't inform the interested reader.

WarrenWeckesser:
B (the table).

mdhaber:
Ok, votes are for the table. I'll try to get this building then merge.

mdhaber:
@ilayn Did I do this right? Didn't seem to work.

mdhaber:
Hey @ilayn thoughts on how to fix this?

ilayn:
Hmm, I guess then it has to go to that `run` block. We just need another line of command to be run and I am not sure if we even need sudo rights.

mdhaber:
@ilayn "Unable to locate package texlive-scientific"
Can we do this without texlive-scientific?

ilayn:
I don't really know where it is. Should be in this https://packages.ubuntu.com/disco/texlive-science but apparently something has changed according to these comments https://askubuntu.com/questions/369268/missing-sty-package. You can remove the S column and make it a regular TeX table if you wish, so that we can merge this.

mdhaber:
@ilayn Can you make the modification and I'll copy-paste?

ilayn:
Ah nice, thank you @mdhaber, @jarrodmillman

PR #119, opened 2019-03-06
Project organization - Governance

mdhaber:
Minor grammatical edits.

If we want to keep all the current content, which is a fine but substantial (~467 word) summary of the governance document and Code of Conduct, I have other suggestions. But do we have space for all of this content?

Here is a reduced version to consider (~215 words).

---
SciPy adopted an official Governance Document on August 3, 2017\cite{SciPyProjectGovernance}.  A Steering Council, currently composed of 18 members, oversees daily development of the project by contributing code and reviewing contributions from the community. Council Members have commit rights to the project repository, but they are expected to merge changes only when there are no substantive community objections. The Chair of the Steering Council, Ralf Gommers, is responsible for initiating biannual technical reviews of project direction and summarizing any private Council activities to the broader community. The project's Benevolent Dictator for Life (BDFL), Pauli Virtanen, has overruling authority on any matter, but the BDFL is expected to act in good faith and only exercise this authority when the Steering Council cannot reach agreement.  

SciPy's official Code of Conduct was approved on October 24, 2017. In summary, there are five Specific Guidelines: 
\emph{be open} to everyone participating in our community; 
\emph{be empathetic and patient} in resolving conflicts; 
\emph{be collaborative}, as we depend on each other to build the library; 
\emph{be inquisitive}, as early identification of issues can prevent serious consequences; and 
\emph{be careful with wording}. 
The Code of Conduct specifies how breaches can be reported to the Code of Conduct Committee, and outlines procedures for the Committee's response. Our Diversity Statement "welcomes and encourages participation by everyone".

PR #121, opened 2019-03-06
Project organization - Governance 2

mdhaber:
As suggested in #119

### DIFF
--- a/scipy-1.0/paper.tex
+++ b/scipy-1.0/paper.tex
@@ -1066,51 +1066,15 @@ inclusion of additional vendored code.}
 
 \subsection*{Governance}
 
-SciPy adopted an official Governance Document on August 3,
-2017\cite{SciPyProjectGovernance}.  We have a Benevolent Dictator for Life
-(BDFL), Pauli Virtanen, leading the project, following a detailed discussion of
-the governance model with the community on the mailing list.  In practice, the
-overruling authority of the BDFL is anticipated to be used only when the
-Steering Council cannot agree on a matter.  The BDFL is expected to consider
-any strong indications that they should step down; although the BDFL may
-appoint a successor, it is generally expected that the Steering Council will be
-consulted on the matter.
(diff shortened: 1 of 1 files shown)


mdhaber:
@rgommers @tylerjereddy I still think something needs to be added to this section so that the distinction between "Council Members" and "maintainers" is clear when the reader gets to the "Maintainers and contributors" section. 
Here:
"A Steering Council, currently composed of 18 members, oversees daily development of the project by contributing code and reviewing contributions from the community".
There:
"The SciPy project has approximately 15 active maintainers---people who review
others' contributions, and in general do everything needed to ensure that the
software and the project move forward"

Perhaps answers to these would help:
- Which of the roles listed in the [governance document](https://docs.scipy.org/doc/scipy/reference/dev/governance/governance.html) do Steering Council Members play most often? Were  "outreach activities" from the original wording particularly important? (The original wording also mentioned that Steering Council Members pick new Steering Council Members, but that did not sound like one of the most important roles to highlight.)
- Is "maintainer" in the paper the same as "Core Developer" in the governance document, or do these sets only partially overlap?
- Are all Steering Council Members also Core Developers?

rgommers:
>  Which of the roles listed in the governance document do Steering Council Members play most often? Were "outreach activities" from the original wording particularly important? (The original wording also mentioned that Steering Council Members pick new Steering Council Members, but that did not sound like one of the most important roles to highlight.)

No, outreach activities are probably not very important. 

> Is "maintainer" in the paper the same as "Core Developer" in the governance document, or do these sets only partially overlap?

Yes, the same. I suggest to only use "maintainer".

> Are all Steering Council Members also Core Developers?

Yes. But not the other way around.

In the text I'd focus on the maintainers, that is the most important group. The Steering Council is a large subset of maintainers (there's some extra requirements to join the steering council, like continued contributions for 1+ year), which as its main role is the body that needs to take decisions that the regular review/maintenance processes cannot make.

PR #123, opened 2019-03-06
Discussion

mdhaber:
We're not allowed to have subheadings in the discussion, so I removed them. I also removed the 20,000 visitors figure from SciPy Today, as the same figure appeared in this section.

There are a few remaining issues. 

1. We mention LIGO here for a second time. It probably needs to be removed.
2. I think the example Cython syntax should be removed. 
3. We can resurrect the fact that the test suite is in good shape but that vendored code could use better test coverage.

Also consider 4) whether we need to rework the discussion section more substantially. The concern is that a discussion should analyze and interpret what has already been presented, whereas right now half of the discussion is a collection of new facts.

In any case, 5) do we also want some sort of Conclusion section? Right now the paper ends abruptly with the example Cython syntax and the roadmap figure.


mdhaber:
OK, I will leave the other things to separate PRs or let them go.

PR #122, opened 2019-03-06
Community beyond

mdhaber:
Minimal changes.
Other suggestions I can put in this PR if they sound like a good idea:

1. The "Governance" and "Maintainers and contributors" subsections are focused on SciPy itself (internal). "Commmunity beyond the SciPy library" is about relationships with other projects (external), but its sandwiched between the other two internal sections. I would suggest making the order "Governance", "Maintainers and contributors", and "Community beyond the SciPy library".
2. This "Project organization and community" section is all about how the SciPy project works. The "Governance" subsection is about the official leadership roles, like the government of a city. The "Maintainers and contributors" subsection is about the people, like the citizens. The "Community beyond the SciPy library" subsection - as written - focuses on relationships with other projects, like neighboring cities. It's much more specific than the title "Community beyond the SciPy library" currently suggests. I think it would make the subsection - and the section as a whole - feel more complete if the title were more specific. Something like "Reciprocal projects"? "Sister projects"? "Symbiotic projects"? 
3. "then returning to the base SciPy library to suggest and even implement improvements _to enable more effective science applications_" could be removed or made more specific to avoid sounding empty. Is this supposed to evoke the thought that the domain-specific libraries want the code they have developed to be available for use in other domains, and they turn to SciPy to make that possible?

PR #128, opened 2019-05-03
Swap subsections in Project Organization

mdhaber:
The "Governance" and "Maintainers and contributors" subsections are focused on SciPy itself (internal). "Commmunity beyond the SciPy library" is about relationships with other projects (external), but its sandwiched between the other two internal sections. I simply changed the order of subsections to "Governance", "Maintainers and contributors", and "Community beyond the SciPy library"; they are now in order from closest to the core of the project to further away.

### DIFF
--- a/scipy-1.0/paper.tex
+++ b/scipy-1.0/paper.tex
@@ -1079,6 +1079,17 @@ SciPy's official Code of Conduct was approved on October 24, 2017. In summary, t
 \emph{be careful with wording}.
 The Code of Conduct specifies how breaches can be reported to a Code of Conduct Committee, and outlines procedures for the Committee's response. Our Diversity Statement ``welcomes and encourages participation by everyone''.
 
+\subsection*{Maintainers and contributors}
+
+The SciPy project has approximately 15 active maintainers---people who review
+others' contributions, and in general do everything needed to ensure that the
+software and the project move forward. Maintainers are critical to the health
+of the project\cite{eghbal2016}; their skills and efforts largely determine how
+fast the project moves forward, and they enable contributions from a much
+larger group. Accordingly, the project also has an additional ~100 unique
+contributors for every 6-monthly release cycle. Anyone with the interest and
+skills can become a contributor or maintainer; the SciPy Developer
+Guide\cite{scipy-dev-guide} provides guidance on how to do that.
 
 \subsection*{Community beyond the SciPy library}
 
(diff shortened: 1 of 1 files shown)


mdhaber:
This was discussed in #122 and there was a +1 from @apbard

scipy-1.0/paper.tex
```diff
@@ -1079,6 +1079,17 @@ \subsection*{Governance}
 \emph{be careful with wording}.
 The Code of Conduct specifies how breaches can be reported to a Code of Conduct Committee, and outlines procedures for the Committee's response. Our Diversity Statement ``welcomes and encourages participation by everyone''.
 
+\subsection*{Maintainers and contributors}
+
+The SciPy project has approximately 15 active maintainers---people who review
+others' contributions, and in general do everything needed to ensure that the
+software and the project move forward. Maintainers are critical to the health
```
ilayn: You can start with the definition of a maintainer and then switch to SciPy specific details. Also approximately 15 is in fact exactly 15 😃

scipy-1.0/paper.tex
```diff
@@ -1079,6 +1079,17 @@ \subsection*{Governance}
...
+of the project\cite{eghbal2016}; their skills and efforts largely determine how
+fast the project moves forward, and they enable contributions from a much
+larger group. Accordingly, the project also has an additional ~100 unique
+contributors for every 6-monthly release cycle. Anyone with the interest and
+skills can become a contributor or maintainer; the SciPy Developer
```
ilayn: I think being a maintainer is a bit more long-term and involved (still of course anyone can become) so we might not want to paint the wrong picture about that here.

mdhaber:
@ilayn thanks! I'll think about incorporating those this weekend, but do you agree with the swap in order? That's all I changed here.

ilayn:
Yes I agree for what its worth. But while we are at it ;-)

scipy-1.0/paper.tex
mdhaber: Is it like “Then shalt thou count to fifteen, no more–no less. Fifteen shall be the number thou shalt count, and the number of the counting shall be fifteen. Sixteen shalt thou not count, neither count thou fourteen, excepting that thou then proceed to fifteen. Seventeen is right out."

But if it might vary, maybe "The SciPy project typically has between 10 and 20 active maintainers..." if we don't like "approximately 15"?

ilayn: Hahah no I mean there are exactly 15 so no need to "roundify" it but OK I take it back :)

mdhaber:
Attempted to address your comments, @ilayn. To do so, I thought it made more sense to swap the order of discussing contributors/maintainers.

ilayn:
Looks good to me.

PR #129, opened 2019-05-03
Technical Improvements punctuation, consistency, convention

mdhaber:
changed "code base" to "codebase" (only) for consistency
"b-spline" to "B-spline according to common convention
Removed ' from ABI's, as some would argue that it is inappropriate
remove inappropriate colon
removed inappropriate comma
Inserted comma after introductory phrase
define COO and DIA


rgommers:
Two approvals, in it goes. Thanks @mdhaber, @apbard, @ilayn

PR #130, opened 2019-05-03
Reductions to paragraphs on local and global optimizers

mdhaber:
As discussed in #113, #102. They needed to be trimmed, and descriptions of old functions needed to be removed.


mdhaber:
@antonior92 and @andyfaff we needed to limit the discussion to recent additions leading to 1.0 and generally cut down many of the technical improvements sections. Are these acceptable?

scipy-1.0/scipy-optimize.tex
```diff
@@ -21,66 +21,8 @@ \subsubsection*{Linear Optimization}
...
...rained optimization were added to \texttt{minimize} in recent versions of SciPy: \texttt{dogleg}, \texttt{trust-ncg}, \texttt{trust-exact}, and \texttt{trust-krylov}. All are trust-region methods that build a local model of the objective function based on first- and second-derivative information, approximate the best point within a local ``trust region'', and iterate until a local minimum of the original objective function is reached, but each has unique characteristics that make it appropriate for certain types of problems. For instance, \texttt{trust-exact} achieves fast convergence by solving the trust-region subproblem almost exactly, but it requires the second-derivative Hessian matrix to be stored and factored every iteration, which may preclude the solution of large problems ($\geq 1000$ variables). On the other hand, \texttt{trust-ncg} and \texttt{trust-krylov} are well-suited to large-scale optimization problems because they do not need to store and factor the Hessian explicitly, instead using second-derivative information in a faster, approximate way. A detailed comparison of the characteristics of all \texttt{minimize} methods is presented in Table~\ref{tab:minimize-si}.
```
apbard:
```suggestion
The \texttt{minimize} function unifies,  in a common interface, several methods for finding local minima of nonlinear optimization problems. Four new methods for unconstrained optimization were added to \texttt{minimize} in recent versions of SciPy: \texttt{dogleg}, \texttt{trust-ncg}, \texttt{trust-exact}, and \texttt{trust-krylov}. All are trust-region methods that build a local model of the objective function based on first and second derivative information, approximate the best point within a local ``trust region'', and iterate until a local minimum of the original objective function is reached, but each has unique characteristics that make it appropriate for certain types of problems. For instance, \texttt{trust-exact} achieves fast convergence by solving the trust-region subproblem almost exactly, but it requires the second-derivative Hessian matrix to be stored and factored every iteration, which may preclude the solution of large problems ($\geq 1000$ variables). On the other hand, \texttt{trust-ncg} and \texttt{trust-krylov} are well-suited to large-scale optimization problems because they do not need to store and factor the Hessian explicitly, instead they use second derivative information in a faster, approximate way. A detailed comparison of the characteristics of all \texttt{minimize} methods is presented in Table~\ref{tab:minimize-si}.
```

scipy-1.0/scipy-optimize.tex
```diff
@@ -21,66 +21,8 @@ \subsubsection*{Linear Optimization}
...
-%This suite is used to decide
-for tracking the performance of existing solvers over time and for evaluating whether the performance of new solvers merits inclusion in the package. 
-% have good enough performance to be added to the package.
+As \texttt{minimize} only seeks a local minimum, some problems require the use of a global optimization routine to find the minimum of the objective function over the set of all decision variable values, especially if the objective function possesses many local minima that other solvers could get stuck in. The new \texttt{scipy.optimize.differential\textunderscore evolution} \cite{Wormington1999,Storn1997} is a stochastic global optimizer that works by evolving a population of candidate solutions. In each iteration, trial candidates are generated by combination of candidates from the existing population. If the trial candidates represent an improvement, then the population is updated. Most recently, the SciPy benchmark suite gained a comprehensive set of 196 global optimization problems for tracking the performance of existing solvers over time and for evaluating whether the performance of new solvers merits inclusion in the package.
```
apbard: If the objective function possesses many local minima, \texttt{minimize} solvers could get stuck into one of those as they only seeks a local minimum. Thus, the use of a global optimization routine to find the minimum of the objective function over the set of all decision variable values is sometimes required. The new \texttt{scipy.optimize.differential\textunderscore evolution} \cite{Wormington1999,Storn1997} is a stochastic global optimizer that works by evolving a population of candidate solutions. In each iteration, trial candidates are generated by combination of candidates from the existing population. If the trial candidates represent an improvement, then the population is updated. Most recently, the SciPy benchmark suite gained a comprehensive set of 196 global optimization problems for tracking the performance of existing solvers over time and for evaluating whether the performance of new solvers worths them the inclusion in the package.

andyfaff:
I wish we were writing using something like Overleaf, the current method by github PR does mess with one's brain. @mdhaber is there any way of looking at the rendered document?

scipy-1.0/scipy-optimize.tex
andyfaff: @apbard, I can't tell the difference between your suggestion and the original.

scipy-1.0/scipy-optimize.tex
andyfaff: Is there a reason why `basinhopping` was removed?

ilayn:
You can check the artifacts tab on CircleCI. https://382-66140682-gh.circle-artifacts.com/0/papers/artifacts/paper.pdf

scipy-1.0/scipy-optimize.tex
apbard: @andyfaff I see...it is rendered as just one line. I'll fix it

apbard:
```diff
-The \texttt{minimize} function unifies several methods of finding local minima 
-of nonlinear optimization problems in a common interface.
+The \texttt{minimize} function unifies, in a common interface, several methods 
+for finding local minima of nonlinear optimization problems. 
 Four new methods for unconstrained optimization were added to \texttt{minimize} 
 in recent versions of SciPy: \texttt{dogleg}, \texttt{trust-ncg}, 
 \texttt{trust-exact}, and \texttt{trust-krylov}. 
 All are trust-region methods that build a local model 
-of the objective function based on first- and second-derivative information, 
+of the objective function based on first and second derivative information, 
 approximate the best point within a local ``trust region'', and iterate until 
 a local minimum of the original objective function is reached, but each has 
 unique characteristics that make it appropriate for certain types of problems. 
 For instance, \texttt{trust-exact} achieves fast convergence by solving 
 the trust-region subproblem almost exactly, but it requires the 
-second-derivative Hessian matrix to be stored and factored every iteration, 
+second derivative Hessian matrix to be stored and factored every iteration, 
 which may preclude the solution of large problems ($\geq 1000$ variables). 
 On the other hand, \texttt{trust-ncg} and \texttt{trust-krylov} are well-suited 
 to large-scale optimization problems because they do not need to store and factor 
-the Hessian explicitly, instead using second-derivative information in a faster, 
+the Hessian explicitly, instead they use second derivative information in a faster, 
 approximate way. A detailed comparison of the characteristics of 
 all \texttt{minimize} methods is presented in Table~\ref{tab:minimize-si}.
```

scipy-1.0/scipy-optimize.tex
apbard:
```diff
-As \texttt{minimize} only seeks a local minimum, some problems require the use of 
-a global optimization routine to find the minimum of the objective function over 
-the set of all decision variable values, especially if the objective function 
-possesses many local minima that other solvers could get stuck in. 
+If the objective function possesses many local minima, \texttt{minimize}'s solvers 
+could get stuck into one of those as they only seeks a local minimum. 
+Thus, the use of a global optimization routine to find the minimum of the 
+objective function over the set of all decision variable values is sometimes required. 
 The new \texttt{scipy.optimize.differential\textunderscore evolution} 
 \cite{Wormington1999,Storn1997} is a stochastic global optimizer that works 
 by evolving a population of candidate solutions. In each iteration, 
trial candidates are generated by combination of candidates from the existing 
population. If the trial candidates represent an improvement, then the population 
is updated. Most recently, the SciPy benchmark suite gained a comprehensive set of 
196 global optimization problems for tracking the performance of existing solvers 
over time and for evaluating whether the performance of new solvers 
-merits inclusion in the package.
+worths them the inclusion in the package.
```

scipy-1.0/scipy-optimize.tex
mdhaber: Thanks for highlighting the changes.

I agree with removing the hyphens. Good catch.

Re: making "in a common interface" a parenthetical clause, I can see arguments both ways.
As the sentence was, "unifies several methods of finding local minima..." was unfortunately separated from "...in a common interface." I would prefer to be able to squash those together. 
Now "unifies, in a common interface, several methods" interrupts the sentence. Is it important enough to interject that thought?

I will fix the hyphens but leave the original "in a common interface", if you don't mind? Of course, if you think it's important enough. please open a PR for it.

scipy-1.0/scipy-optimize.tex
mdhaber: The "Key technical improvements" section is supposed to "describe key technical improvements made in the last three years" (prior to 1.0 branch). `basinhopping` is much older than that. Was it substantially approved within the window?

mdhaber: According to MW, use of "worths" as a verb is archaic.
https://www.merriam-webster.com/dictionary/worth
It was also an intransitive verb, so I don't think this would have been a proper way to use it, anyway. Also, I don't think the definite article "the" is appropriate for "inclusion".

I think your point is that we're talking about including the "new solvers" in the package, not including their "performance" in the package?
In that case, I can change it to "merits _their_ inclusion in the package."

mdhaber: Re: first sentence, was the reason for the change to break up the long sentence? I can see that.

I'd prefer to avoid the passive "is sometimes required", though. And I never thought the phrase "get stuck in" was really appropriate here; "get stuck into one of those" is less so.

If the problem was mostly that the sentence was too long, how about just cutting out some extraneous bits?

"As \texttt{minimize} may return any local minimum, some problems require the use of 
a global optimization routine."

scipy-1.0/scipy-optimize.tex
apbard: I also changed the `several methods of finding local` to  `several methods `**for**` finding local`, isn't this more correct?

Do you think something like `The \texttt{minimize} provides a common (or shared) interface for several methods of... ` solves the issue? If not, your version it is fine.

mdhaber: Yes, both of these are good. I'll fix.

mdhaber: I went with "unified" interface. Means the same thing but suggests that they might have been separate before, and I think that's why the original author wrote something like "unifies ... in a common interface."

scipy-1.0/scipy-optimize.tex
mdhaber: @andyfaff Look OK?

scipy-1.0/scipy-optimize.tex
```diff
@@ -21,66 +21,8 @@ \subsubsection*{Linear Optimization}
...
-%  \label{eq:minimization-prob}
-%  \min_x f(x)
-%\end{equation}
-%where $f$ is a multivariate function $f: \mathbb{R}^n \rightarrow \mathbb{R}$ and $x \in \mathbb{R}^n$.
-%%  Without the problem statement this sentence isn't really appropriate.
-%\texttt{COBYLA} and \texttt{SLSQP} methods support nonlinear constraints, \texttt{L-BFGS-B} and \texttt{TNC} allow box constraints $x^l \le x \le x^u$,
-%and the remaining methods are for unconstrained problems.
-
-% The available methods implement a variety of solution strategies.
-Derivative-free methods (e.g. \texttt{Nelder-Mead}, \texttt{Powell} and \texttt{COBYLA}) are appropriate for minimizing non-differentiable or
```
andyfaff: It's a shame to lose lines 42 to 49. But if there's no space...

scipy-1.0/scipy-optimize.tex
```diff
@@ -21,66 +21,8 @@ \subsubsection*{Linear Optimization}
...
-% They're ``comprehensive''
-% of varying difficulties, 
-% that can be used to evaluate the performance of the solvers. 
-%This suite is used to decide
-for tracking the performance of existing solvers over time and for evaluating whether the performance of new solvers merits inclusion in the package. 
-% have good enough performance to be added to the package.
+As \texttt{minimize} may return any local minimum, some problems require the use of a global optimization routine. The new \texttt{scipy.optimize.differential\textunderscore evolution} \cite{Wormington1999,Storn1997} is a stochastic global optimizer that works by evolving a population of candidate solutions. In each iteration, trial candidates are generated by combination of candidates from the existing population. If the trial candidates represent an improvement, then the population is updated. Most recently, the SciPy benchmark suite gained a comprehensive set of 196 global optimization problems for tracking the performance of existing solvers over time and for evaluating whether the performance of new solvers merits their inclusion in the package.
```
andyfaff: 'The new diffev *function*  is a ...'

andyfaff: Does the table of minimizers include basin_hopping/brute, etc? If not then I think we should put them in there, they're a critical part of optimize. If Nelder-Mead belongs in the table, so do they.

scipy-1.0/scipy-optimize.tex
mdhaber: Those derivative free solvers are old, right? If not <3 years old, it's out of scope.
It's good to concisely give context for the recent stuff, and that's why I kept the bit explaining the trust region approach. But all of the new stuff is trust region and derivative-based, so I don't think we can describe all the other types of optimization approaches.

scipy-1.0/scipy-optimize.tex
mdhaber: @antonior92 put that table together for the local minimizers if I remember correctly. You're welcome to add the global ones, I think.

mdhaber: @andyfaff This text uses `scipy.optimize.differential_evolution` as a noun.  "`scipy.optimize.differential_evolution` function" would be using `scipy.optimize.differential_evolution` as a noun adjunct. I think either is acceptable and it's really a matter of preference, but if you'd like I can change it.

mdhaber:
Approved by +2. Merging.

PR #131, opened 2019-05-03
Mention what cKDTree does

mdhaber:
The section about cKDTree is quite long, so I thought it would be worth
mentioning what k-d trees are.


rgommers:
makes sense, merged - thanks!

PR #132, opened 2019-05-03
Trim LowLevelCallable

mdhaber:
As discussed in #113


rgommers:
That still captures the essence, merging. thanks!

PR #133, opened 2019-05-03
Sparse / cKDTree tweaks

mdhaber:
There were duplicate definitions of CSR/CSC format acronyms in adjacent sections sparse and cKDTree. The sparse section seemed a more natural place to define them than right before it in cKDTree, so I swapped the order and removed the duplicate definitions from cKDTree.

(Note: COO and DIA definitions added to sparse in #129)


rgommers:
Merged, thanks @mdhaber

PR #134, opened 2019-05-03
Community beyond the SciPy library section

apbard:
Starting from discussion on #122

As far as I can see, the message this paragraph is trying to convey is that the communities of downstream projects actively contribute to scipy and by doing so indirectly contibute to other projects. Thus, scipy plays a mediator and meeting point role in the python ecosystem.
 
I don't dislike the current title, but I see @mdhaber point about more specific title and I don't mind changing it. What about "Relation with sibiling projects"?

### DIFF
--- a/scipy-1.0/paper.tex
+++ b/scipy-1.0/paper.tex
@@ -1088,7 +1088,7 @@ For example, there are common contributors to the SciPy and
 Astropy core libraries\cite{astropy-2018}, and what works 
 well for one of the codebases, infrastructures, or communities
 is often transferred in some form to the other. At the codebase
-level, the \texttt{binned_statistic} functionality
+level, the \texttt{binned\_statistic} functionality
 is one such cross-project contribution: it was initially
 developed in an Astropy-affiliated package
 and then placed in SciPy afterwards. 
--- a/scipy-1.0/paper.tex
+++ b/scipy-1.0/paper.tex
@@ -1081,18 +1081,19 @@ The Code of Conduct specifies how breaches can be reported to a Code of Conduct 
 \subsection*{Downstream projects}
 
 The scientific Python ecosystem includes many examples
-of domain-specific software libraries that are built 
-on top of SciPy features. The work done on these libraries
-often turns out to be beneficial to the whole ecosystem
-being SciPy the catalyst of this cross-fertilisation process. 
+of domain-specific software libraries building on top
+of SciPy features and then returning to the base SciPy library
+to suggest and even implement improvements. 
 For example, there are common contributors to the SciPy and 
 Astropy core libraries\cite{astropy-2018}, and what works 
-well for one of the codebases, infrastructures, or communities 
-is often transferred in some form to the other. At the codebase 
-level, the \texttt{binned\_statistic} functionality
-is one such cross-project contribution---it was initially
+well for one of the codebases, infrastructures, or communities
+is often transferred in some form to the other. At the codebase
+level, the \texttt{binned_statistic} functionality
+is one such cross-project contribution: it was initially
 developed in an Astropy-affiliated package
-and then placed in SciPy afterwards.
+and then placed in SciPy afterwards. 
+In this perspective, SciPy serves as a catalyst for cross-fertilisation 
+throughout the Python scientific computing community.
 
 \subsection*{Maintainers and contributors}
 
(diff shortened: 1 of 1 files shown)


ilayn:
Why don't you use the word "downstream" already? It is the right terminology.

apbard:
> 
> 
> Why don't you use the word "downstream" already? It is the right terminology.

You mean "Relation with downstream projects"? If yes, you are right. definetly better than sibiling

ilayn:
Yes, exactly. Sorry for the terse comment.

scipy-1.0/paper.tex
```diff
@@ -1078,16 +1078,16 @@ \subsection*{Governance}
...
-of SciPy features, and then returning to the base SciPy library
-to suggest and even implement improvements to enable
-more effective science applications. For example, there
-are common contributors to the SciPy and Astropy core
-libraries\cite{astropy-2018}, and what works well for 
-one of the codebases, infrastructures, or communities 
+of domain-specific software libraries that are built 
+on top of SciPy features. The work done on these libraries
+often turns out to be beneficial to the whole ecosystem
+being Scipy the catalyst of this cross-fertilisation process.
```
mdhaber: SciPy needs to be stylized consistently with the P capitalized.
Is "the whole ecosystem being SciPy the catalyst..." grammatically correct?

Also, I'm not sure that the rewording is 100% improved.

On one hand, it removes the phrase "to enable more effective science applications", which sounded empty to me. I don't _disagree_ with the statement at all, but think we already established that "improvements [to SciPy]" enable more effective science applications. Stating claims explicitly like this can make people question them more. There are examples of this in the news all the time!

On the other hand, the first sentence (in isolation) has less meaning now: "other libraries build on top of SciPy". That, too, has already been established in the paper. I think it's better to keep the idea that "downstream libraries contribute back to SciPy" in the same sentence (as in the original).

scipy-1.0/paper.tex
```diff
@@ -1078,16 +1078,16 @@ \subsection*{Governance}
 The Code of Conduct specifies how breaches can be reported to a Code of Conduct Committee, and outlines procedures for the Committee's response. Our Diversity Statement ``welcomes and encourages participation by everyone''.
 
 
-\subsection*{Community beyond the SciPy library}
+\subsection*{Relation with downstream projects}
```
mdhaber: How about just "Downstream Projects"? 

The section is called "Project organization and community", and I think what we're trying to say with these subsection headings is that the project is organized into three groups: 

- "Governance", in the sense of "the persons/committees /departments who make up a body for the purpose of administering something",
- "Maintainers and Contributors", and
- "Downstream Projects".

"Relation with downstream projects" is not a group (of people) like the first two are.

scipy-1.0/paper.tex
apbard: > SciPy needs to be stylized consistently with the P capitalized.
Fixed

> Is "the whole ecosystem being SciPy the catalyst..." grammatically correct?
I believe so, but I am not a native speaker so I maybe wrong.

Any suggestion about how to improve the rest?

scipy-1.0/paper.tex
apbard: Done

scipy-1.0/paper.tex
mdhaber: Well, I would prefer to just remove the phrase "to enable more effective science application". I didn't have a problem with the rest of the original.

What were you trying to improve, or what did you not like about the original? Maybe we can find something that sounds better to both of us.

apbard: I was trying to say this:

> As far as I can see, the message this paragraph is trying to convey is that the communities of downstream projects actively contribute to scipy and by doing so indirectly contibute to other projects. Thus, scipy plays a mediator and meeting point role in the python ecosystem.

that is, a little step more than just say that "downstream libraries contribute back to SciPy".

mdhaber: OK, I think that's a good point to add. I'll think about it some more.

mdhaber:
@apbard I think your point is great, but rather than reworking the whole paragraph, I'd prefer to point it out at the end. How about:

The scientific Python ecosystem includes many examples
of domain-specific software libraries building on top
of SciPy features and then returning to the base SciPy library
to suggest and even implement improvements. For example, there
are common contributors to the SciPy and Astropy core
libraries\cite{astropy-2018}, and what works well for 
one of the codebases, infrastructures, or communities
is often transferred in some form to the other. At the codebase 
level, the \texttt{binned\_statistic} functionality
is one such cross-project contribution: it was initially
developed in an Astropy-affiliated package
and then placed in SciPy afterwards. **In this way, SciPy serves 
as a catalyst for cross-fertilisation throughout the Python scientific 
computing community.**

apbard:
@mdhaber I like your suggestion but I think we should either put it just before the example or change `In this way` to something else like `from this perspective`?

apbard:
@mdhaber fixed

mdhaber:
Oops. Should have squash merged, probably. Worth reverting?

ilayn:
For the article I don't think it matters much

PR #140, opened 2019-05-16
Attempt to make use of (sub)package/(sub)module consistent

### LINKED ISSUE #47: figure out the right wording for (sub)package & (sub)module)
We now use subpackage, package, module and submodule very inconsistently in various parts of the paper. Let's go through once we have completed most/all sections and do this right.

mdhaber:
Key word being "attempt". I'm not sure if all of these changes are correct, but this will hopefully at least highlight potential misuses of the these words.
Closes #47

### DIFF
--- a/scipy-1.0/ckdtree.tex
+++ b/scipy-1.0/ckdtree.tex
@@ -1,4 +1,4 @@
-The \texttt{scipy.spatial.cKDTree} subpackage, which implements a space-partitioning data structure that
+The \texttt{scipy.spatial.ckdtree} module, which implements a space-partitioning data structure that
 organizes points in $k$-dimensional space, was rewritten in C++ with templated classes. 
 Support was added for periodic boundary conditions, which are often used 
 in simulations of physical processes. 
--- a/scipy-1.0/ckdtree.tex
+++ b/scipy-1.0/ckdtree.tex
@@ -1,4 +1,4 @@
-The \texttt{cKDTree} module, which implements a space-partitioning data structure that
+The \texttt{scipy.spatial.cKDTree} subpackage, which implements a space-partitioning data structure that
 organizes points in $k$-dimensional space, was rewritten in C++ with templated classes. 
 Support was added for periodic boundary conditions, which are often used 
 in simulations of physical processes. 
--- a/scipy-1.0/paper.tex
+++ b/scipy-1.0/paper.tex
@@ -173,7 +173,7 @@ around the world. For example, published
 scripts\cite{alex_nitz_2018_1596771, LIGO-open}
 used in the analysis of gravitational waves
 \cite{PhysRevD.93.122003, abbott2017gw170817} 
-import several submodules of SciPy, and the M87 black
+import several subpackages of SciPy, and the M87 black
 hole imaging project directly cites SciPy\cite{2019ApJ875L3E}.
 
 Recently, SciPy released version 1.0, a milestone that traditionally
@@ -813,8 +813,8 @@ changes that improve clarity in the API of the library but strongly discourage
 breaking backwards compatibility given our position near the base of the 
 scientific Python computing stack.
 
-In addition to the Python API, SciPy has C and Cython interfaces in a number
-of submodules. Therefore, we have to also consider the application binary
+In addition to the Python API, SciPy has C and Cython interfaces.
+Therefore, we have to also consider the application binary
 interface (ABI). This ABI has been stable for a long time, and we aim to
 evolve it only in a backwards compatible way.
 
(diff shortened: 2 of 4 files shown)


scipy-1.0/subpackages.tex
```diff
@@ -47,20 +45,20 @@
...
-    This package contains various functions for multi-dimensional image
+    This subpackage contains various functions for multi-dimensional image
     processing, including convolution and assorted linear and nonlinear
     filters (Gaussian filter, median filter, Sobel filter, etc.);
     interpolation; region labeling and processing; and mathematical morphology
     functions.
 \item[\texttt{misc}]
     A collection of functions that did not fit into the other subpackages.
     While this subpackage still exists in SciPy 1.0, effort is underway
-    to deprecate or relocate the contents of this subpackage and eventually remove it.
```
mdhaber: I removed this word purely for formatting: "it" was by itself on the next line.  I don't think this is inaccurate. Work is underway to deprecate/relocate its contents and remove it.

scipy-1.0/paper.tex
```diff
@@ -872,8 +872,8 @@ \subsection*{Cython bindings for BLAS, LAPACK, and special}
 LAPACK\cite{LAPACK} routines for many years. SciPy now additionally
 includes Cython\cite{behnel2011cython} wrappers for
 many BLAS and LAPACK routines (added in 2015) and the special functions 
-provided in the \texttt{scipy.{\allowbreak}special} submodule (added in 2016).  
-These Cython wrappers are available in the modules
+provided in the \texttt{scipy.{\allowbreak}special} subpackage (added in 2016).
```
tylerjereddy: That looks ok I think:

```
In [8]: help(special)
Help on package scipy.special in scipy:

NAME
    scipy.special
```

But the official docs so allow use of (sub) prefix for these of course.

scipy-1.0/ckdtree.tex
```diff
@@ -1,4 +1,4 @@
-The \texttt{cKDTree} module, which implements a space-partitioning data structure that
+The \texttt{scipy.spatial.cKDTree} subpackage, which implements a space-partitioning data structure that
```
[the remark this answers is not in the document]
mdhaber: According to the interpreter, `scipy` is a module, too.
```
import scipy
print(repr(scipy))

```
But that is not consistent with the definition of a module provided in the [documentation](https://docs.python.org/3/tutorial/modules.html). 

"A module is a file containing Python definitions and statements."

(@rgommers seemed to prefer this strict definition in #47 if I understood correctly.)

I checked before committing, and cKDTree isn't a file.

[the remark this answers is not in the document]
mdhaber: But yes, I'll change the capitalization. Thanks for catching that.

scipy-1.0/subpackages.tex
```diff
@@ -10,9 +10,7 @@
 \begin{description}[leftmargin=!, labelwidth=\widthof{\bfseries \texttt{interpolate}}]
 \itemsep0em
 \item[\texttt{cluster}]
-    The \texttt{cluster} subpackage contains two modules:
-    \texttt{cluster.vq} provides vector quantization and $k$-means algorithms,
-    and \texttt{cluster.hierarchy} provides functions for hierarchical and
+    The \texttt{cluster} subpackage contains \texttt{cluster.vq}, which provides vector quantization and $k$-means algorithms, and \texttt{cluster.hierarchy}, which provides functions for hierarchical and
```
mdhaber: These are indeed modules. Should I revert?

scipy-1.0/subpackages.tex
```diff
@@ -73,23 +71,19 @@
...
+		decomposition and LU factorization.
     \texttt{scipy.sparse.csgraph} provides a collections of graph algorithms
     for which the graph is represented using a sparse matrix.  Algorithms
     include connected components, shortest path, minimum spanning tree
     and more.
 \item[\texttt{spatial}]
     This subpackage provides spatial data structures and algorithms,
     including the $k$-d tree, Delaunay triangulation, convex hulls and Voronoi
-    diagrams.  The subpackage \texttt{scipy.spatial.distance} provides
+    diagrams. \texttt{scipy.spatial.distance} provides
```
mdhaber: This is actually a module, too.

tylerjereddy: I think your decision to remove the specification here is just fine, so I'll mark as resolved.

scipy-1.0/subpackages.tex
tylerjereddy: Nah, what you did looks fine to me. Completely avoiding the confusion seems nice too.

tylerjereddy:
Ok, minor revisions were made--in it goes.

The Python [glossary](https://docs.python.org/3/glossary.html#term-package) makes the situation even more complicated to think about:

> package
> A Python module which can contain submodules or recursively, subpackages. Technically, a package is a Python module with an __path__ attribute.

Of course, SciPy compiles extension modules from Cython and so on, too. I suspect there could be a fairly long debate about some individual cases, but I'm happy for the issue to be considered closed! They are things available in the SciPy namespace!

The glossary also defines `namespace package` from PEP420, which serves only as a container for subpackages, but have no physical representation...

tylerjereddy:
thanks Matt

PR #144, opened 2019-05-19
MAINT: add benchmark supporting data

tylerjereddy:
I'll set this as a draft PR and update as the benchmark results come in by amending / force pushing. Most likely not suitable to store the data assets under version control, though maybe a single tarball when they're all done might not be too bad. Otherwise, figshare or whatever.

We'll have to see how they're looking with a different machine this time. Not sure how much of a pain it is to parse the JSON data files, many of which will contain failures for historical dependency reasons, but presumably there are good Python libraries for scraping out the data.

* add JSON benchmark data files for cKDTree query()

* add text file containing git diff / patch for improving
the historical sampling of the pertinent benchmark in spatial


tylerjereddy:
The [data should be regenerated now](https://workflowy.com/s/i-need-to-regenerate/yug1dImVfF6C8FTu) in this PR at path `scipy-1.0/supporting_info/asv_bench/cKDTree` and in particular the machine subfolder `adobe/`.

That said, there's about 30 MB of raw JSON data files in here, so regeneration was just part of the challenge and next up is mining the data out.

I've not been able to use the usual `asv` mechansims to preview the results first. I think Pauli mentioned somewhere something about symlinking the results directories back to the repo of the project proper, but basically I think we'll end up having to parse through those JSON files with i.e., Python to pull out the data for a custom plot.

If someone does start working on that:

1. Let us know so I don't duplicate effort--otherwise, I will probably try to start on it sometime today.
2. The asv internals probably provide useful clues about parsing the JSON
3. Remember Cython 0.27.3, 0.21.1 and 0.18 were used for three different passes through the benchmarks. Files should be named accordingly, along with whether I used `conda` or `virtualenv` (it varied).

mdhaber:
If the json is what it was like from asv of unladen SciPy, I already have a head start. I'll take a look today.

mdhaber:
Looks doable. I'll try to update the figure this afternoon/weekend.

mdhaber:
@tylerjereddy Any chance we're missing data from commits c. mid 2015- early 2018? I think I'm processing the data correctly but here's the rough plot:
rough
That last commit in 2015 is March 24 (19e0b8ab3e192cfb438348d33b004018685d39d4), the one date in 2018 is March 5 (b3ddb2c8a21c8c89bd7474ec40bdb045977a311a), and there doesn't seem to be anything in between.
Let me know if I need to double-check.

mdhaber:
Another weird thing is that around the time of the major jumps in performance there is some "overlap"; that is, there are some commits after the performance first improves that still have the old performance. Can you interpret that? I think it makes sense, but I wonder why it was not seen in the old plot?

mdhaber:
Regarding the plot style, I think I will only have three colors: one per version of Cython. The three different `m` values form distinct traces; I can just label each of them.

tylerjereddy:
The original plot terminates at SciPy 0.18 (around September 2016 or so), but the data you show is still cut off too early even for that.

It does look like one of the plateaus is missing relative to the previous plot, though it seems like a small time gap that is missing given above constraint. Though indeed, one might argue getting to 1.0 might make sense on the plot, nobody has complained about that time span in last 12+ months.

My notes linked above indicate that for the most recent commits (Cython 0.27.3 builds) it was necessary to transition from using conda to virtualenv for asv benchmarks, so this may produce a ton of decoy / failure files with conda-based filenames, but there should be decent data in the virtualenv-named versions, I hope. Could that throw off your script?

tylerjereddy:
For the issue of performance lag--not exactly sure, but non-linear commit history / merging / branching behavior seems like a plausible consideration to start with. Not sure if `asv` does anything fancy to make that a little nicer for sorting commits.

mdhaber:
@tylerjereddy I don't use the filenames at all. All the data about the Cython version, etc., is inside each file.

Re: performance lag, that's what I thought. I wonder what `asv` does to avoid showing that. 

I double-checked by simply recording for each file the timestamp, commit, and whether the file contained data. None of the files contain data during that span. Here are the beginning of those filenames.

```['01529b', '02a4b9', '02e978', '058202', '0c42ad', '0d4341', '0eb827', '0f7383', '113589', '14ac4f', '15183d', '17325e', '198208', '1a8772', '1c8385', '1c8759', '1d0140', '1d3529', '1ef67e', '1f72ee', '1fdb99', '2dc406', '2de476', '331e2c', '33b68b', '37b6a0', '38e24c', '396a35', '3af64e', '3bb02c', '3deede', '3e2764', '3f2308', '4019b4', '41b413', '439cbc', '46cbf5', '4b6543', '4d59b1', '4ea460', '4f4046', '5070ab', '511d92', '519d3c', '59cabc', '59d444', '615be9', '615fd6', '633e86', '6439e8', '654219', '65901f', '66e7ff', '688407', '691a12', '69742b', '6c5c19', '6d310f', '6f0f1b', '6f205e', '6fd2bc', '705b70', '70897a', '7498c3', '76cee1', '770cf1', '7ac143', '7af8e8', '808013', '8346ee', '845e46', '87a6cf', '8b2cd5', '921f56', '95f480', '97e2a9', '9bc28e', '9c4475', 'a13695', 'a46a15', 'a81bc7', 'a8248a', 'a890b7', 'a8ade2', 'ab146c', 'ab4d82', 'ab5ef3', 'abf764', 'b15282', 'b1f2ef', 'b36adf', 'b39e62', 'b4070f', 'b43ea2', 'b60f38', 'b6736a', 'b86506', 'ba79a1', 'bb3c6b', 'bb8bbe', 'bca1d9', 'bca265', 'bcd070', 'bd0876', 'bfce07', 'c0bb90', 'c96c52', 'c9dd9e', 'cc7bf1', 'cc9f6b', 'ccb338', 'cce8e3', 'cd64e3', 'ce4d74', 'cf1610', 'd03d1f', 'd1ee8d', 'd1ffd3', 'd26941', 'd3f28b', 'd4be9d', 'd6e61e', 'd784c3', 'd9ff7c', 'daf498', 'deb4e9', 'df0c7c', 'dfc8d5', 'e3a2f9', 'e499e5', 'e6ffa7', 'eae2ea', 'f06b7e', 'f22851', 'f3c1df', 'f59cc2', 'f7784f', 'f79762', 'fc302b', 'ff9ca1']```
 
Those files have `"results"->"spatial.Query.time_query"->"result"` `null`, indicating a failed benchmark.

**Comment by mdhaber** [comment]
If it helps to have a commit range, the last good commit before the gap is:
`a1bfecb17170541101c06b41f0035f9e7b5b01ff`
The first commit in the gap is:
`c0bb9080e28620806c3f0582277831b897847ae2`
The last commit in the gap is:
`2de47611aff67e78c8d7ba6f1a893cfc6ac8801b`
And there is one commit after that for which there is some data:
`b3ddb2c8a21c8c89bd7474ec40bdb045977a311a`

**Comment by tylerjereddy** [comment]
picking out `69742b` (v0.17pre):

- all three Cython version JSON files do indeed record null results
- I did keep two of the 100 MB log files for the entire benchmark runs, and for this commit they say:
  - `log_cython_018.txt`: `Building SciPy requires Cython >= 0.22` (okay, not helpful)
  - `log_cython_0273.txt`: `cc1: fatal error: scipy/sparse/linalg/dsolve/SuperLU/SRC/sgsisx.c: No such file or directory` (that's maybe a bit strange?)

**Comment by tylerjereddy** [comment]
Not sure that's a good enough clue for me to think of something offhand I can change to fix it & start it running before the long weekend. 

My usual approach would be to set up a docker container to match that config / commit hash as closely as possible, figure out what needs to change to remedy the error, and then do an asv pass with that change. So, that's likely what I'd have to do I guess.

**Comment by mdhaber** [comment]
That would be appreciated. The plot would look twice as interesting with a second jump in performance.

**Comment by tylerjereddy** [comment]
This shows up for a lot of commits:

```
File "/mnt/c/Users/treddy/github_projects/scipy/benchmarks/env/1d8282ebbb739c5f0fe84ba5c9fdcbed/lib/python2.7/site-packages/numpy/f2py/crackfortran.py", line 833, in markoutercomma
    assert not f, repr((f, line, l))
AssertionError: (-1, 'intent(in))', 'intent(in))')
```

Ideally there'd be a better way to provide system isolation to avoid polluting the benchmark environments--it looks like even virtualenv is still relatively easy to pollute with i.e., f2py from somewhere else. An added complication may be virtualenv benchmarks invoked from within an exterior conda env & the way that numpy-distutils handles this at the 1.8.2 version point.

**Comment by tylerjereddy** [comment]
Ah, some of my issues have been related to this: https://github.com/airspeed-velocity/asv/issues/822

Rolling back to a less bleeding-edge version of asv is probably warranted for the remaining benchmarks.

**Comment by tylerjereddy** [comment]
@mdhaber I've synced in the current progress with Cython 0.27.3 + asv 0.3, which should be slowly filling in the remaining gaps. I'll do one more sync when that run finishes--it says about 13 % progress for this pass, but realistically it is usually more than that because many historic commits will just hard fail / skip with modern Cython.

Hopefully this starts looking better now.

**Comment by mdhaber** [comment]
Yup, this appears to be filling in the gap:

Figure_1

**Comment by tylerjereddy** [comment]
I've synced in the remaining Cython 0.27.3 + asv 0.3 results. That pass crashed out near SciPy 0.9.0, but I suspect that's far enough back to finish up.

If there's anything else we need for the purposes of reproducibility / plot aesthetics now is the time to ask. I think there are detailed links to what I've done, but it can be hard to put it all together. Basically mix of conda / virtualenv in the `asv.conf.json`, asv 0.4 dev branch & asv 0.3 stable as runners, with a variety of `asv.conf.json` files used committed in this PR branch.

The other thing is the time lags discussed above--I suppose we could `git grep` around in `asv` to figure out how they "linearize" the history a little better if we really want.

**Comment by mdhaber** [comment]
hmmm.

Figure_1

For the top trace (biggest problem, longest runs), the latest run did a great job of filling in the gaps until the end of 2017, which is all we want.

But for the other two traces, the execution times are substantially different with the new version of Cython. 

It looks like the method of timing the tests has changed between the versions of ASV. The precision of the timing appeared to be lot finer in the older data (5/24) taken with the "bleeding edge" version. For example, commit 7ac86ab1 was tested successfully with all three versions of Cython with the original version of CSV being used. Here are the times for the first few tests in Cython 0.18 (presumably bleeding edge ASV, if I understand correctly):
```
                0.0016593592507498606, 
                0.003050476312637329, 
                0.001650639942714146, 
                0.0030597448348999023, 
                0.0016552720751081195, 
```
For Cython 0021.1:
```
                0.0016435725348336355, 
                0.0030345916748046875, 
                0.0016496522086007254, 
                0.0030271410942077637, 
                0.0016582182475498744, 
```
For Cython 0.27.3:
```
                0.001667158944266183, 
                0.003057509660720825, 
                0.0016498565673828125, 
                0.003075122833251953, 
                0.001671433448791504, 
```
But for Cython 0.27.3 + ASV 0.3:
```
                0.0026041666666666665, 
                0.00390625, 
                0.0026041666666666665, 
                0.00390625, 
                0.0026041666666666665, 
```
Does the old ASV use a different clock?

pv:
If you're on windows, you should upgrade to asv 0.4.1.

tylerjereddy:
All benchmarks were executed on Windows Subsystem for Linux, and yes with different versions of asv because of the pip forced reinstall bug causing havoc on the bleeding edge.

I could try to repeat with asv 0.4.1 for Cython 0.27.3 and sync that set of results in a separate commit--is that what you want?

mdhaber:
I don't want you to have to repeat it, but it looks like it's necessary to get usable data for this benchmark.
@rgommers would you agree?

tylerjereddy:
Ok, I've restarted with `asv 0.4.1` and Cython 0.27.3; I'll let that run over the weekend

tylerjereddy:
Ok, I've synced in the results for asv 0.4.1 + Cython 0.27.3 in a new commit. Not sure if you'll need to revert the penultimate commit with the asv 0.3 data, but hopefully this helps.

mdhaber:
It looks like it worked. It looks like it overwrote all the files, so the plot looks good:
Figure_1
I think all asv 0.3 data has been overwritten or replaced, but if you can confirm by actually reverting the commit I would appreciate it. I can't seem to figure out how to revert it properly.

tylerjereddy:
Ah yeah, reverting is a nightmare here because of all the data files getting overwritten periodically I think & there's more than 1 commit with Cython 0.27.3 + asv 0.3. `git revert HASH` doesn't work so well with all the modifications.

I'm not too bothered if we don't merge this--the main thing is to get the data files to you & perhaps some modest updates if needed for reproducibility. If we do want the data in the SI in this repo, we could try a tarball instead of 3202 JSON files taking up 34 MB in folder `adobe`.

Could also use figshare or something like that.

tylerjereddy:
Feel free to close this and just open a PR with a tarball or put the stuff on figshare. Or I can do that later this evening / tomorrow.

mdhaber:
That's fine as long as you think _all_ the files were overwritten. If there were some files that were not overwritten, then those will appear as spurious data points from ASV 0.3 that shouldn't appear with the rest of the points.

tylerjereddy:
I can't guarantee that every single file was overwritten, but the information is likely possible to dissect out with some pain. If someone wants to step up and parse that out they are more than free to do so--otherwise, a small disclaimer in the manuscript might suffice.

There should be more than enough detail now to reproduce what I've done if there's still desire by someone to do another pass to produce "pure" data.

There might be something to be said for having the JSON files tagged with asv version used to produce them, but I think there are also possibly some pains with trying to change the format at this point.

mdhaber:
Used git to record filenames in the commits and compared the sets. Looks like 274 files from the 5/30 commit were not modified today. I can remove any of them that actually had data from the plot.

rgommers:
@tylerjereddy should we merge this, or leave the PR open indefinitely?

tylerjereddy:
@mdhaber Do you have a preference on how we should make this data available? Merging is easiest, but pretty ugly on the size & mess it creates in the repo. Lots of options like figshare I suppose..

mdhaber:
The nice thing about putting it in the repo is that it makes the Data Availability statement easy. But otherwise no preference. Just have to update the Data Availability statement accordingly.

rgommers:
I didn't check the size and number of files. Anything up to 10 MB is a no-brainer, just merge it here. In the 10-100 MB range it becomes a bit more tricky. @tylerjereddy could you make a recommendation?

tylerjereddy:
Originally this was > 3000 text files and 35 MB of data.

I've now compressed the majority of the text files into `ckdtree_asv_supporting.tar.xz`, which is 3 MB.

There are 4 other small text files included--felt safer not deleting them.

Anyway, I rewrote the history of this branch to only add 5 files and 3 MB of data, so should be good to go with a simple merge now?

rgommers:
Sounds good, in it goes. Thanks Tyler!

ilayn:
I just made a simple CSV file from the data @mdhaber provided kindly. It's a few kB for the benchmark figure. We can also include it in the file. I'll put the result in #173

PR #146, opened 2019-06-03
Integrated @WarrenWeckesser ending from #9

mdhaber:
Thanks @WarrenWeckesser for the concluding remarks in #9. I hope this PR integrates these as intended.


mdhaber:
I think it is important to read the conclusion as it appears in the document:
https://445-66140682-gh.circle-artifacts.com/0/papers/artifacts/paper.pdf

I think it wraps things up a lot better than the snippets of Cython code. But although these are actually legitimate observations, I'm concerned that it may come across to reviewers as self-congratulatory, as the people named in this section are authors : /

Thoughts?

WarrenWeckesser:
After I wrote those remarks and let them sit for a while, I wondered if anyone that I named explicitly would rather not be called out like that, and if it would be better without giving specific names. Instead we could refer to something like "a small contingent of developers that have been contributing 10 years or more".  I'm fine with removing the names.  The point is that scipy is a very active project that continues to grow and improve, driven by a development team that includes a few members who have been around 10 or more years, a good number (yeah, that's vague, I know) who have been involved 5+ years, and a bunch of new folks.

Matt (or anyone else), feel free to edit the comments into any form that seems more suitable for the paper.

Also, I think the comments about wanting more tutorials and cleaning up the cython code could be removed completely.   But I won't insist if there is a strong preference for keeping them.

rkern:
As one so named, I don't mind one way or the other. :-)

rgommers:
I don't mind being named personally either way. By the way @WarrenWeckesser, I consider summer 2008 my starting point - I wrote a ton of docs in both doc marathons:) 

Trying to look through the eyes of a reader or reviewer: yes I think it's better to remove the names. The ending is much better than before, I like it. The message won't change with dropping the names though, and that paragraph will be more compact. Let's try to edit that down to the essentials:
- problem statement
- our mix of people/roles
- healthy outlook for SciPy

stefanv:
Either approach is fine with me; and I fully intend to write code for SciPy again ;)

mdhaber:
@rgommers @WarrenWeckesser Removed shout-outs and some details, preserving most of the language:
https://447-66140682-gh.circle-artifacts.com/0/papers/artifacts/paper.pdf

WarrenWeckesser:
Looks good.

rgommers:
LGTM too, merged. Thanks all!

PR #147, opened 2019-06-06
Replaced benchmarking figure/caption

### LINKED ISSUE #143: ASV Benchmarking Figure
There have been concerns about the quality of the ASV Benchmarking Figure (#65, #117).

asv_time_query_ckdtree

Apparently, the data are no longer available and are difficult to generate, so I scraped the data from [here](https://pv.github.io/scipy-bench/#spatial.CNeighbors.time_count_neighbors_deep?y-axis-scale=log&p-Nr=10&p-Nr=100&p-Nr=1000&p-(m%2C%20n1%2C%20n2)=(2%2C%201000%2C%201000)&p-(m%2C%20n1%2C%20n2)=(8%2C%201000%2C%201000)&p-(m%2C%20n1%2C%20n2)=(16%2C%201000%2C%201000)&x-axis-scale=date%2Finfo.json&Cython=0.27.3&python=2.7&numpy=1.13.3) for `spatial.CNeighbors.time_count_neighbors_deep` benchmark and produced:

asv

If you have style tweak suggestions that would be easy for me to implement (in Inkscape), I will consider them. I would welcome suggestions (as RGBA hex codes) for the color palette. I would also welcome a draft caption, as I don't know what this benchmark really does.

I will submit a PR to replace the figure after I find out what the improvements were. If you happen to have ideas about those, please let me know. In case it helps, the dates and commit hashes corresponding with each data point are [here](https://www.dropbox.com/s/enw9hy6g9v81344/asvdata.csv?dl=0).

I have no attachment to this benchmark and I'd prefer to have a denser set of points, so if the data for the original benchmark is found or recreated I'd be happy to swap it in.

mdhaber:
Closes #143


mdhaber:
The horizontal lines look a little dark on screen but print nicely.
https://446-66140682-gh.circle-artifacts.com/0/papers/artifacts/paper.pdf

ilayn:
If you have the data we can draw it via `pgfplots` with all kinds of aesthetic shananigans needed. So +1 from me to merge.

rgommers:
This vertical line is still clearly off:

image

Either you have the wrong commit, or the improvement came from an algorithmic change after the C++ rewrite. Would be good to move the line and figure out if the current statement that it's the C++ rewrite is correct.

mdhaber:
@rgommers The line passes through the first three points for which the time has dropped. I suppose I'll have too make those bigger than the rest.
We can discuss this on the call.

rgommers:
> The line passes through the first three points for which the time has dropped. I suppose I'll have too make those bigger than the rest.

That's a single commit, could be an outlier in benchmarking or could be that a commit from that PR was picked up by asv before the PR was merged. The line should go through where the average level changes, ~20 points to the right.

mdhaber:
> That's a single commit, could be an outlier in benchmarking 
The line for "`cKDTree` fully Cythonized" marks a commit in which all 12 tests represented on the plot (different _m_s, different versions of Cython) were completed in about half the time compared to the corresponding tests in previous commits. The magnitude of the drop in execution time is far beyond that of all previous outliers, and these lower times are consistent with the times in future commits. The probability of this occurring by chance seems low. The data is more easily explained by something in the code changing. The same is true for the commit marked  "`cKDTree` rewritten in C++".

>or could be that a commit from that PR was picked up by asv before the PR was merged
Yes, seems more likely. 

> The line should go through where the average level changes, 
It is not so obvious to me that it is the right thing to do, but I will create a version like that.

tylerjereddy:
Seems likely that big performance-enhancing PRs get merged with some old commits because they took so long to develop & review, while simple / faster-to-develop PRs get merged just before with newer commit hashes.

So commits with more recent dates lack the performance enhancements, because timestamp and merge order are not the same thing.

A->B->C happened on master, but A->C->B happened if you only filter by timestamp.

The lag times are usually not super long because people will eventually start rebasing before merges I think.

I think `asv` may have a few tricks for cleaning this up a little in its own plots. Looking around in the source I see stuff like `git rev-list --first-parent`, but there seem to be tons of `git` options about ordering log output in any case, if that's really something we're worried about.

Remember that the original plot I provided was not by date, but by commit, and using `git log` or similar under the hood with the right options might have provided a less confusing feature-ordered independent variable instead of hard-enforcing chronological ordering on i.e., timestamps?

tylerjereddy:
Another idea is just including merge commits, which should be properly feature-ordered I think (?), but that may cause too much sparsity without i.e, another re-run.

mdhaber:
> The line should go through where the average level changes,
_Now_ it's clear that this is the right thing to do based on the investigation reported in #143. 

Originally, I thought these individual commits where the performance first improved were part of the series of commits that made the performance improvements. They are not. For whatever reason their parents are commits that are months in the future, long after the performance improvements were made.  I removed the points corresponding with these commits because they are confusing and really aren't important.

mdhaber:
@tylerjereddy I just saw your comments. Naw, I don't think we need to do all that.

mdhaber:
@ilayn 
> If you have the data we can draw it via pgfplots with all kinds of aesthetic shananigans needed.

I'm trying not to be offended? What aesthetic shenanigans are needed?

ilayn:
Ah sorry if it come out as dismissive, it is quite the opposite. 

What I meant is that we can TeXify the plots so that they match the fonts surrounding them and the line grayness etc. just as we did for other plots. That has two benefits (no extra figure file and directly modifiable in the TeX source code). I think this is already very nice so no complaints 😃

mdhaber:
@ilayn the fonts should already match. I generated all the text with LaTeX. I do see that if everything were redone in pgfplots it would make it easier to keep styles consistent. Thanks!

ilayn:
Yes, indeed. That's why I mentioned about the availability of the data. 

I am a frequent user of it so exporting to pgfplots goes very quickly for me but as you say, I don't want to step on anybody's toes about it.

PR #148, opened 2019-06-07
Pruned end material

mdhaber:
Per @rgommers in #65. See comments in self-review.
Uncertain what to remove v.s. keep. Easiest if we remove everything, as none of this is information the paper can't do without. If not, what are the criteria for being included? There were a lot of people who wrote material that was cut. Can it all be shifted here, or how should it be selected?


scipy-1.0/paper.tex
```diff
@@ -1156,25 +1157,17 @@ \section*{Discussion}
 
 \bibliography{references}
 
-\section*{Acknowledgments}
```
mdhaber: Removed per @rgommers comment in #65

scipy-1.0/paper.tex
```diff
@@ -1156,25 +1157,17 @@ \section*{Discussion}
...
 \section*{Author contributions statement}
 
-\fixme{TODO}
+M.H. edited the manuscript, created figures, compiled the author list and contributions statements, and drafted the \texttt{linprog} technical improvements section.
 
 %Must include all authors, identified by initials, for example: A.A.
 %conceived the experiment(s),  A.A. and B.A. conducted the experiment(s), C.A.
 %and D.A. analysed the results.  All authors reviewed the manuscript.
 
-\section*{Additional information}
```
mdhaber: Unused, so removed.

scipy-1.0/paper.tex
```diff
@@ -1184,28 +1177,30 @@ \section*{Supporting Information}
 \setcounter{figure}{0}
 \setcounter{table}{0}
 
-\begin{figure}[H]
```
mdhaber: Not certain that this needs to be removed entirely, but should it be a whole page? Suggest to match style with other figures.

scipy-1.0/paper.tex
```diff
@@ -1184,28 +1177,30 @@ \section*{Supporting Information}
...
-\centering
-\includegraphics[width=0.8\textwidth]{supporting_info/knn_complexity_confirm}
-\caption{Assessment of \texttt{cKDTree} nearest neighbor query time complexity as a function of the number of input points (N) along with similar assessment for related algorithms in scikit-learn. The calculations are based on code previously reported\cite{knn-jake}, but updated to use SciPy 1.0.0, Python 3.6, NumPy 1.13.1, and Cython 0.27.3.}
-\label{fig:knn-complexity}
-\end{figure}
 
-\begin{figure}[H]
```
mdhaber: Can't include screenshots, see #143.

scipy-1.0/paper.tex
```diff
@@ -1246,7 +1241,7 @@ \section{Polynomial Interpolators: Additional Details}
 \code{BSpline} objects instead of directly manipulating \code{tck}
 tuples of knots, coefficients and the spline degree.
 
-\section*{Numerical Optimization: Additional Details}
+\subsection*{Numerical Optimization: Additional Details}
```
mdhaber: @rgommers suggested moving this to front, but this causes some page layout issues. Maybe send it for review looking better like this, mentioning to reviewers that it will be moved to front before submission?

scipy-1.0/paper.tex
```diff
@@ -1184,28 +1177,30 @@ \section*{Supporting Information}
...
+%\end{figure}
+%
+%\begin{figure}[H]
+%\centering
+%\includegraphics[width=\textwidth]{supporting_info/asv_bench/sparse/sparse_coo_dia_to_csr_csc}
+%\caption{Airspeed velocity benchmarks for SciPy COO / DIA to CSR / CSC matrix format conversion over roughly three years of commits prior to the release of version 1.0. Python version 2.7, Cython version 0.27.3, and NumPy 1.8.2 were used. The values were obtained from our regularly-updated suite of performance results available online at \url{http://pv.github.io/scipy-bench/}.}
+%\label{fig:sparse-conv}
+%\end{figure}
+
+\subsection*{Polynomial Interpolators: Additional Details}
```
mdhaber: Should this be kept?

rgommers: Actually not too useful probably, could be cut.

rgommers:
> Uncertain what to remove v.s. keep. Easiest if we remove everything, ...

Good question. I think the one really nice bit of content is the table of optimizers. The paper says:
_A detailed comparison of the characteristics of all minimize methods is presented in Table S1_
We could just move the table there, and then delete the rest of the SI. The text could be changed to something like:
__A detailed comparison of the characteristics of all `minimize` methods is presented in Table 1. It illustrates the level of completeness that SciPy aims for when covering a numerical method or topic._ (note it's early, the sentence is not great, hope you get the idea).

scipy-1.0/paper.tex
tylerjereddy: Pretty sure you can keep screenshots in SI. The standards for SI are generally far less strict that for the main body of the manuscript, including formatting related matters.

tylerjereddy:
> none of this is information the paper can't do without

That's the point of Supporting Information--otherwise the reviewers will say that it has to be moved to the main body of the manuscript. 

I don't understand why we need to prune the SI--it has 4 pages only at the moment, which is nothing compared to a big scientific paper SI. It doesn't have to be perfect or super well-formatted compared to the main manuscript & can just serve to store some extra useful but non-essential information. I'd just move it to a separate PDF for the submission process & move on.

scipy-1.0/paper.tex
```diff
@@ -1156,25 +1157,17 @@ \section*{Discussion}
...
-\fixme{TODO}
-
-%Acknowledgements should be brief, and should not include thanks to anonymous
-%referees and editors, or effusive comments. Grant or contribution numbers may
-%be acknowledged.
-
 \section*{Author contributions statement}
 
-\fixme{TODO}
+M.H. edited the manuscript, created figures, compiled the author list and contributions statements, and drafted the \texttt{linprog} technical improvements section.
```
tylerjereddy: So we're going to do this systematically then?

mdhaber: If you're referring to the level of detail -systematically summarizing all contributions, I'm not sure. I thought I should give it a go so we could decide here.

If you are referring me compiling a list, it says in the template that all the authors have to be in the contributions statement. If they really mean that everyone's initials have to appear, I thought it would be easier for someone to collect and add them. I thought I'd just include it on the author form I made last year. Everyone who has contributed (for which an email address is available) will get an email with their PR/issue contributions, and they could summarize concisely?

Should probably include a bit about my contribution to SciPy itself rather than just the paper?

mdhaber:
Re: SI - if it's in a separate PDF, I wouldn't mind including it all in the submission. I think then we could even throw some other stuff in that authors submitted but weren't included. I'm not sure about stripping it down just to the optimize table. Seems inconsistent to let that stay and everything else go.

scipy-1.0/paper.tex
rgommers: I'd suggest to take a shortcut here, to avoid a large amount of work. What the AstroPy paper did: list separately all project maintainers, the people who did most of the writing, and all other contributors (on page 1). Then here you can simply say "group X did ..., group Y did ...".

rgommers:
> I'm not sure about stripping it down just to the optimize table. Seems inconsistent to let that stay and everything else go.

Yes that would look odd, that's why I suggested to put it in the paper itself. 

There's two arguments for cutting the rest:
1. least amount of work (because it's not ready as is)
2. none of it is very important.

The `cKDTree` figure looks nice, the rest is all first draft quality it looks like to me.

mdhaber:
BTW here is more information about SI:
https://www.nature.com/srep/publish/guidelines#supplementary-info

I see I also need to add Additional Information back for a "Competing Interests" statement.

WarrenWeckesser:
Regarding the SI: I think we can drop it completely with no loss to the paper.  The material is interesting to *us*, the SciPy developers, but to most readers of this paper, it will be uninteresting and, for the most part, quickly irrelevant.  Some of those details can be just as easily found in the online SciPy documentation, and that will be up-to-date.  The table of the optimization methods is very nice, but not essential.

I read through the paper a few days ago, and it looks very good.  There are a few very small changes I'd make, but they are at the level of copy-editing, not major restructuring or rewriting.  The point of the paper is to say, in effect, "SciPy 1.0 exists!  Here's some history, some recent developments, and the current status", and it does that just fine.

I suggest dropping the SI, do the final copy-editing of the text, and call it done.  It's time to be ruthless.

mdhaber:
Followed @rgommers @WarrenWeckesser advice: moved `minimize` methods table to main text and removed remaining SI material. Added other required sections.

After merge, should I send a link to the core-dev list first (before contacting all potential authors)?

mdhaber:
BTW regarding the markers for contributing material/editing, I intended to add those after we have the author list.

scipy-1.0/subpackages.tex
```diff
@@ -59,7 +59,7 @@
     Orthogonal distance regression, including Python wrappers for the Fortran
     library ODRPACK\cite{ODRPACK_Boggs}.
 \item[\texttt{optimize}]
-    This subpackage includes the following, with additional details in the SI:
+    This subpackage includes the following, with additional details in Supporting Information:
```
tylerjereddy: This isn't true anymore right?

scipy-1.0/scipy-optimize.tex
```diff
@@ -21,8 +21,65 @@ \subsubsection*{Linear Optimization}
...
+      \cite{powell_direct_1994, powell_direct_1998, powell_view_2007} &
+      \cite{polak_note_1969, nocedal_numerical_2006} & \cite{nocedal_numerical_2006} & \cite{byrd_limited_1995, zhu_algorithm_1997} &
+      \cite{schittkowski_nonlinear_1982, schittkowski_nonlinear_1982-1, schittkowski_convergence_1983, kraft_software_1988} &
+      \cite{nash_newton-type_1984} & \cite{nocedal_numerical_2006}  & 
+      \cite{powell_new_1970, nocedal_numerical_2006} &  \cite{steihaug_conjugate_1983, nocedal_numerical_2006} &
+      \cite{conn_trust_2000, more_computing_1983} & \cite{gould_solving_1999, lenders_trlib:_2016} \\
+    \hline
+  \end{tabular}
+  \label{tab:minimize-si}
+\end{table}
```
tylerjereddy: Didn't we already have an extended discussion about where this table belongs--I thought it started in the main paper, and then went to the SI, and now it is back.

I guess people just weren't monitoring closely X months ago.

scipy-1.0/subpackages.tex
mdhaber: Nope, good catch.

scipy-1.0/scipy-optimize.tex
mdhaber: Well, it made sense at the time. The same arguments for moving it out of the main text still apply, but now the arguments for moving it back in are stronger. Go iterative methods.

tylerjereddy:
Alright, in it goes by majority

PR #151, opened 2019-06-19
Grammar and other nit picks

jnothman:
I've not finished reading the paper yet, but here are some nit picks to get started.

### DIFF
--- a/scipy-1.0/ckdtree.tex
+++ b/scipy-1.0/ckdtree.tex
@@ -15,8 +15,8 @@ while preserving the asymptotic complexity.
 
 In 2015, SciPy added the \texttt{sparse\_distance\_matrix} routine for generating 
 approximate sparse distance matrices between \texttt{KDTree} objects by ignoring 
-all distances that exceed a user-provided value. Also, this routine is not 
-limited to the conventional L2 (Euclidean) norm but supports any Minkowski 
+all distances that exceed a user-provided value. This routine is not 
+limited to the conventional L2 (Euclidean) norm, but supports any Minkowski 
 $p$-norm between 1 and infinity. By default, the returned data structure is a
 Dictionary Of Keys (DOK) based sparse matrix, which is very efficient for matrix 
 construction. This hashing approach to sparse matrix assembly can be 7 times 
--- a/scipy-1.0/ckdtree.tex
+++ b/scipy-1.0/ckdtree.tex
@@ -16,7 +16,7 @@ while preserving the asymptotic complexity.
 In 2015, SciPy added the \texttt{sparse\_distance\_matrix} routine for generating 
 approximate sparse distance matrices between \texttt{KDTree} objects by ignoring 
 all distances that exceed a user-provided value. This routine is not 
-limited to the conventional L2 (Euclidean) norm, but supports any Minkowski 
+limited to the conventional L2 (Euclidean) norm but supports any Minkowski 
 $p$-norm between 1 and infinity. By default, the returned data structure is a
 Dictionary Of Keys (DOK) based sparse matrix, which is very efficient for matrix 
 construction. This hashing approach to sparse matrix assembly can be 7 times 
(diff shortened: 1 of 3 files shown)


scipy-1.0/paper.tex
```diff
@@ -693,11 +693,11 @@ \subsection*{Project scope}
 PyMC3\cite{10.7717/peerj-cs.55}, emcee\cite{2013PASP-emcee} and 
 PyStan\cite{pystan-ref} cover Bayesian statistics and probabilistic modeling.
 scikit-image\cite{vanderwalt2014scikit} provides image processing
-capabilities beyond \texttt{scipy.ndimage}, sympy\cite{meurer2017sympy}
```
jnothman: the notation `scipy.x` is not introduced, while my proposed change here is more consistent with the reference to `sparse` and `spatial` later in the sentence

scipy-1.0/ckdtree.tex
```diff
@@ -15,8 +15,8 @@
 
 In 2015, SciPy added the \texttt{sparse\_distance\_matrix} routine for generating 
 approximate sparse distance matrices between \texttt{KDTree} objects by ignoring 
-all distances that exceed a user-provided value. Also, this routine is not 
-limited to the conventional L2 (Euclidean) norm but supports any Minkowski 
+all distances that exceed a user-provided value. This routine is not 
+limited to the conventional L2 (Euclidean) norm, but supports any Minkowski
```
mdhaber: Removing "Also" is fine, but this new comma is not appropriate because "supports any Minkowski $p$-norm between 1 and infinity" is not an independent clause. 
https://www.grammarly.com/blog/comma-before-but/

scipy-1.0/paper.tex
mdhaber: Changing to SciPy's \texttt{ndimage} makes sense if we do not expect the reader to understand the subpackage notation. I'll leave that to others to decide.

scipy-1.0/paper.tex
```diff
@@ -693,11 +693,11 @@ \subsection*{Project scope}
...
+capabilities beyond SciPy's \texttt{ndimage}, sympy\cite{meurer2017sympy}
 provides a Python interface for symbolic computation, and
 NetworkX\cite{hagberg2008networkx} is a Python library for probing
 graphs and networks in a more specialized way than SciPy components
-like \texttt{sparse.csgraph} or \texttt{spatial}.
+like \texttt{sparse.csgraph} or \texttt{spatial} offer.
```
mdhaber: Consider the simplified sentence:
"NetworkX is a Python library for probing graphs in a more specialized way than sparse.csgraph."
vs 
"NetworkX is a Python library for probing graphs in a more specialized way than sparse.csgraph offers."
I'm sure that the first version can be improved, but I'm not convinced that appending "offers" is the way to do it.

scipy-1.0/paper.tex
```diff
@@ -711,7 +711,7 @@ \subsection*{Project scope}
 In terms of software systems and architecture, SciPy's scope matches NumPy's:
 algorithms for in-memory computing on single machines, with support for a wide
 range of data types and process architectures. Distributed computing and support
-for graphics processing units (GPUs) is explicitly out of scope.
+for graphics processing units (GPUs) are explicitly out of scope.
```
mdhaber: Agreed.

scipy-1.0/paper.tex
jnothman: Ambivalent

scipy-1.0/ckdtree.tex
jnothman: Grammarly doesn't provide an example comparable to this where the entire verb phrase is coordinated by but. I would go for whichever reads better over a textbook prescription here. Which do you think reads better?

jnothman: Someone else commented on the grammarly blog about the absence of this verb phrase coordination case. I've adopted your/their preference in any case.

jnothman:
I've gone through the paper (but not references). I'm unlikely to add to this more here.

As a general comment, it might be nice, given the venue, to make a note of how the scientific community should/could be contributing to scipy and its roadmap.

jnothman:
PS: seeing as you've invited several people recently, it would be good to merge these minor changes quickly before everyone conflicts with  their nit picking.

rgommers:
LGTM, merged. Thanks @jnothman

PR #152, opened 2019-06-19
Minor language tweaks

jni:
Thanks for a wonderfully written paper! I learned a couple of things even though I try to read the release notes quite carefully! =)

The "seemed to" that I removed below should at least be changed to "seems to", but I think it's best to just remove it altogether.

Additionally, I noticed that the number of dependent repos and packages are different between the abstract and the main text (under SciPy Today). I presume the main text numbers are correct, as the abstract numbers seem very low for SciPy.

### DIFF
--- a/scipy-1.0/paper.tex
+++ b/scipy-1.0/paper.tex
@@ -192,9 +192,9 @@ ecosystem of research software---backed by millions in funding and
 many hundreds of highly qualified engineers\cite{mathworks-globe-97,
 esri-revenue,bloom-wolfram}---was preposterous.
 
-Yet, the philosophical motivations behind a fully open toolstack, and
-an excited, friendly community with a singular focus, have
-proven auspicious in the long run.  It led not only to the library
+Yet the philosophical motivations behind a fully open toolstack, combined
+with an excited, friendly community with a singular focus, have
+proven auspicious in the long run.  They led not only to the library
 described in this paper, but also to an entire ecosystem of related
 packages \cite{scipy-ecosystem}, and a variety of social activities centered around
 them\cite{social-python}. The packages in the SciPy ecosystem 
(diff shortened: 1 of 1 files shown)


scipy-1.0/paper.tex
```diff
@@ -193,7 +193,7 @@ \section*{Introduction}
 esri-revenue,bloom-wolfram}---was preposterous.
 
 Yet, the philosophical motivations behind a fully open toolstack, and
-an excited, friendly community with a singular focus, seemed to have
+an excited, friendly community with a singular focus, have
```
mdhaber: I agree with change. But now that you highlight this sentence, the original comma usage is probably wrong, so it would be helpful if you could fix that in this PR. 
Specifically, the comma after yet seems inappropriate: https://english.stackexchange.com/questions/106901/yet-at-the-beginning-of-a-sentence
And the commas surrounding ", and an excited, friendly community with a singular focus," would only be appropriate if this is a parenthetical clause, which I don't think it is. 
The fact that:
"Yet the philosophical motivations behind a fully open toolstack and an excited, friendly community with a singular focus have proven auspicious in the long run."
looks a bit strange does not mean that the commas were used correctly in the first place. It might indicate that the sentence needs to be revised more substantially, though.

Would you mind researching this and either remove the offending commas or argue to keep them? Thank you!

scipy-1.0/paper.tex
```diff
@@ -257,7 +257,7 @@ \section*{Background}
 features that one would expect to find in a general-purpose
 programming language.
 \end{quote}
-This is still a distinguishing feature of Python for science, and one of the
+This remains a distinguishing feature of Python for science, and one of the
```
mdhaber: OK.

scipy-1.0/paper.tex
jni: @mdhaber very good, fixed! As a minor grammar nazi, I appreciate the prod. =P

mdhaber:
Thanks @jni!

PR #154, opened 2019-06-21
AUTH: add affiliation

andyfaff:


mdhaber:
In general let's do this in bulk when everyone has filled out the author form. But this is fine so I'll merge it.

PR #156, opened 2019-06-21
add missing "from" and "."

rc:
After careful reading, I have found only two little omissions - great work!

There is some trailing white-space in the .tex files - is it worth "fixing"?


rgommers:
thanks @rc 

I think we probably shouldn't bother with whitespace fixes for now

PR #160, opened 2019-06-25
Added my name to authors list

ilayn:
Sometimes academic customs change and last author means a different thing. So please let me know if this is the wrong place, I have absolutely no preference about the order.


rgommers:
thanks Ilhan. let's try and merge these PRs quickly as they come in, to avoid merge conflicts. We'll deal with author name ordering at the end.

PR #168, opened 2019-06-28
Moved all figures and tables to end

### LINKED ISSUE #164: Figure Legends
Figure legends are limited to 350 words.
"References should be numbered sequentially, first throughout the text, then in tables, followed by figures; that is, references that only appear in tables or figures should be last in the reference list."
Maybe nothing needs to be changed, but we need to check this carefully.

### LINKED ISSUE #166: Move Tables to End of Document
"Please submit tables at the end of your text document (in Word or TeX/LaTeX, as appropriate)."

https://www.nature.com/srep/publish/checklist

mdhaber:
Closes #164.
Closes #166.


tylerjereddy:
I don't think we should move figures to the end--it helps readability to have them inline at appropriate locations and is allowed for PDF submissions as far as I can tell.

mdhaber:
You're right, "PDF" is conspicuously absent from the rule:
Please submit tables at the end of your text document (in Word or TeX/LaTeX, as appropriate).

Then again, I also thought this was an easy way of getting the reference numbers to satisfy the rule: "References should be numbered sequentially, first throughout the text, then in tables, followed by figures;"

Fine with me if the consensus is to close without merging; just trying to check all the boxes.

rgommers:
> I don't think we should move figures to the end--it helps readability to have them inline at appropriate locations and is allowed for PDF submissions as far as I can tell.

I agree, I hate reading draft articles with figures all at the end. Seems to be allowed as is, so let's keep it at that.

PR #182, opened 2019-07-07
Correct my affiliation (I'm based in Montreal, not Cambridge)

fabianp:
It seems that two researchers from Google but different offices are authors in this paper, and I was erroneously attached to the Cambridge address. I took the approach of creating a new entry for me, changing the Cambridge address to Montreal. It would also be fine for me to remove geographical information and have as affiliation simply Google LLC for both me and Paul van Mulbregt (but I would need to check with Paul first).

Also corrected Google Inc. (old affiliation) to Google LLC


mdhaber:
Note: fixed on Google Sheet

fabianp:
Thanks!

PR #187, opened 2019-07-19
Fix non-ASCII last name

Juanlu001:
In case I'm still in time!


tylerjereddy:
Oy, the git diff !

Probably not too late for an eventual fix, although I can't actually see the fix easily. I believe Matt generates the author list programmatically from form data.

mdhaber:
@Juanlu001 please let us know what needs to be changed. We will not merge this; it just needs to be adjusted in our spreadsheet.

ilayn:
Do you mind making the list newline broken when repopulating it? TeX doesn't care and will make things a bit more visible. 

Nano-nitpick: Two letter TeX commands are deprecated long time ago. Either use `\textbf{...}` for short items or `{\bfseries ... }` for the long groups, especially important if you are going to mix other font shapes.

Juanlu001:
Oops, didn't realize the diff...

It's `Rodr\'iguez$^{70}`, my second last name, instead of `Rodr'iguez$^{70}`, which is what I typed in the form.

mdhaber:
@Juanlu001 all set

mdhaber:
@ilayn next version will have line breaks and `bfseries`

PR #190, opened 2019-07-21
SciPy provides sparse arrays -> matrices

mdhaber:
Wouldn't that be nice!


ilayn:
This distinction will not be apparent to the reader though since they are kind of interchangeable in the theoretical sense.

PR #192, opened 2019-07-21
Cleaned references more

mdhaber:
Hopefully we won't be rejected based on the reference section being messy : )


rgommers:
thanks Matt

PR #199, opened 2019-07-26
Fix typo in Special Functions

mirca:
I believe this is a typo.  I'm unaware of "hanker" function and I haven't found anything in the SciPy documentation.  I think the original writer meant "Hankel" function.  Apologies if I'm mistaken.


mirca:
Oh I'm sorry, I just read the caption of the excerpt.

PR #200, opened 2019-07-26
Fixes typo (hanker => Hankel)

arokem:


endolith:
I made the same mistake, and so did @mirca https://github.com/scipy/scipy-articles/pull/199

It's probably best to fix the typos (maybe inside [brackets]) and replace ` (typos included)` with ` (typos corrected in brackets)`  (`Netwon-CG` is the other one)

tylerjereddy:
@endolith are you saying this isn't ready to merge yet?

apbard:
@tylerjereddy this shouldn't be merged like this. The typo, as stated in the caption, was present in the original release note that is reported "as it is".
If understand correctly, @endolith is suggesting that maybe is better to correct the typos and state in the caption that the release note has been fixed.

arokem:
Sorry about that. How about a comment on that line, just to make sure that no one else does this?

mdhaber:
@tylerjereddy unless you feel strongly either way I suggest merging with only the comments added; the journal can advise.

endolith:
Yeah, I think it's better to fix the typos so the meaning is clear, since we're in a better place to understand what was actually meant than the reader is

mdhaber:
@endolith That's not what I meant, but if you prefer, please indicate your support of #203

PR #193, opened 2019-07-21
MAINT: adjust covlett for Nature Methods

tylerjereddy:
* adjust the cover letter in preparation
for submission to *Nature Methods* after
receiving Editorial bounce-back from
*Scientific Reports* ("out of scope")

* also included are formatting changes
used for the previous submission cover letter,
so that it fits on a single page,
but phone numbers have again been censored
to protect PII

From the [instructions for *Nature Methods*](https://www.nature.com/nmeth/for-authors/preparing-your-submission), Cover Letter section:

> You must disclose details of any related manuscripts that you have under consideration or in press elsewhere

> You should indicate whether you have had any prior discussions with a Nature Methods editor about the work described in the manuscript.

Those should be addressed now--Editorial Contact is via @stefanv.

Here's the [aims & scope page](https://www.nature.com/nmeth//about/aims). We perhaps need to be cautious of the bio focus, but it is certainly true that we meet "Articles are selected for publication that present broad interest."

PR #201, opened 2019-07-26
MAINT: Fix Bezier spelling

MartinThoma:
* Make sure line breaks / spaces are set correctly
  - See https://tex.stackexchange.com/a/22563/5645
* Use \num from siunitx for number formattingi
* Use \textelp{} from csquotes
* Remove trailing spaces


MartinThoma:
If you really want `,` instead of a thinspace as a thousands seperator, I suggest to use

```
\sisetup{group-separator = {,}}
```

scipy-1.0/covlett.tex
```diff
@@ -56,8 +56,8 @@
 
 We are submitting the enclosed manuscript describing a milestone
 1.0 release of SciPy for consideration to be published in
-\emph{Scientific Reports}. SciPy is the fundamental scientific 
-computing library in the Python programming language. 
+\emph{Scientific Reports}. SciPy is the fundamental scientific
+computing library in the Python programming language.
```
tylerjereddy: Probably don't need to clean up the cover letter since #193 will adjust it for a new journal anyway & it is not really part of any source submission, I don't think.

mdhaber:
@MartinThoma Don't need ",". SI prefers space anyway.

 `\num` changes the font or size of the `\num{100000}` in the abstract. What to do about that?

MartinThoma:
> \num changes the font or size of the \num{100000} in the abstract. What to do about that?

Good catch! I fixed it (`\sisetup{detect-all}`)

mdhaber:
What is the `checkwriting` file? Do we need it?

MartinThoma:
> What is the `checkwriting` file? Do we need it?

Ooops, that was a mistake. [Academic-Writing-Check](https://github.com/devd/Academic-Writing-Check) is a tool I use for my publications. No, it should not be in here. I removed it.

mdhaber:
OK. Thanks @MartinThoma

PR #202, opened 2019-07-26
Change location of MPI-SWS.

musically-ut:
Correct the city for MPI-SWS.


mdhaber:
I changed this on the spreadsheet from which the author list is generated, so closing here.

PR #203, opened 2019-07-26
Fix SciPy 1.0 announcement typos

mdhaber:
Adding this alternative to move things along. Thanks for bringing it up @arokem.

Supersedes and Closes #200 

I'd suggest _not_ adding brackets, as the source is cited and it's just not important enough to call attention to the typos. But I will change if majority vote insists.

@endolith, @apbard, @tylerjereddy, @arokem please case your votes

### DIFF
--- a/scipy-1.0/paper.tex
+++ b/scipy-1.0/paper.tex
@@ -301,19 +301,19 @@ SciPy is an open source package that builds on the strengths of Python and
 Numeric providing a wide range of fast scientific and numeric functionality.
 SciPy's current module set includes the following:
 
-    Special Functions (Bessel, hanker, Airy, etc.)
+    Special Functions (Bessel, Hankel, Airy, etc.) % hanker -> Hankel
     Signal/Image Processing
     2D Plotting capabilities
     Integration
     ODE solvers
-    Optimization (simplex, BFGS, Netwon-CG, etc.)
+    Optimization (simplex, BFGS, Newton-CG, etc.) % Netwon -> Newton
     Genetic Algorithms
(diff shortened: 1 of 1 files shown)


apbard:
No strong opinion. Whatever you choose is fine.

endolith:
I'm fine with no brackets

tylerjereddy (approved):
Looks like we're ok to merge here based on feedback / majority.

tylerjereddy:
@rgommers Pointed out that the `%` comments here actually made it into the final article PDF: https://rdcu.be/b08Wh

I suppose that is a bit unusual vs. using `[]` corrections for quotes in prose. Do we want to follow-up on that? Not sure much can be done after proofing/online publication.

The PDF also only gives one of the three corresponding authors an asterisk, though I wasn't going to bother bringing that up since the html is ok for that.

mdhaber:
The arXiv version looks like:
image

That is, typos are included, and we indicate that in the caption.

Based on the comments in gh-200, the purpose of this PR was to simply correct the typos and indicate in the caption that we did so. None of the comments following `%` were suppose to appear.

I think it is worth mentioning to the journal that comments in Box 1 following `%`  were not supposed to be printed. Those comments indicate that the original message to the mailing list had typos, and the intent was for the journal to choose whether to print with the typos included or with the typos corrected (potentially indicating what was corrected).

Also, it's unclear where the capitalization first appeared, but Table 2 subpackage names were not supposed to be capitalized.

andyfaff:
differentialevolution should be differential_evolution

PR #211, opened 2019-07-28
Fixed spatial description in project scope

mdhaber:
Closes #210 .
@charris ?
The affiliation will be added next time I generate the author list.

PR #213, opened 2019-07-28
Address #209

### LINKED ISSUE #209: Suggestions for further improvement
First of all, I like this paper a lot. It is really very interesting and very well written. Thanks to everybody who has contributed to it.

Apart from a few smaller fixes collected in #208, I have a few suggestions for (small) changes.

* On page 3, "to solve and minimize nonlinear equations" does not make sense to me because equations cannot be minimized. An option could be "to solve nonlinear equations and minimize nonlinear functions" even though I admit that the text becomes more clumsy.

* On page 4, it may not be clear to every reader what event is referred to in sentence 3 of the third paragraph in the section "SciPy matures". After all, the previous paragraph talks about documentation. How about replacing "The event" by "The SciPy conference"?

* On page 7, the last couple of lines in the first paragraph of the section "Project scope" seem broken. I do not think that a construction "while ... compared to ..." is possible. A simple solution could be to drop "while".

* In table 1, the only line in the table not referenced in the legend (apart from the references which are self-explained) is "Line search (LS) or trust-region (TR)". This line might deserve being mentioned in the legend.

mdhaber:
@gertingold I agree with your points in #209. The third is addressed in #211. Do these changes address your concerns?

If so,
Closes #209

### DIFF
--- a/scipy-1.0/paper.tex
+++ b/scipy-1.0/paper.tex
@@ -273,8 +273,9 @@ package, and provided algorithms for signal processing, special
 functions, sparse matrices, quadrature, optimization, Fast Fourier
 Transforms, and more.  One of these packages, Multipack\cite{multipack}, was a set of
 extension modules that wrapped Fortran and C libraries such as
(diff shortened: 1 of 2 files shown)


gertingold:
Yes, this addresses all my points. Thanks, @mdhaber, for your work.

Issue #214, opened 2019-07-28
References in Table 1


FabricioS:
The references in the last line of the Table 1 could be more nicely formatted using the \citen command from the cite package (already loaded by the wlscirep class currently used).

Direct use makes the table wider, and scaling could be appropriate
`\newcommand{\inlinecite}[1]{\footnotesize\citen{#1}}`
in header, and
`\inlinecite{nelder_simplex_1965, wright_direct_1996}`
at https://github.com/scipy/scipy-articles/blob/91afe6546f6b7caf37fe0f9fe8200a0ea3ee535f/scipy-1.0/scipy-optimize.tex#L73

mdhaber:
Please open a PR so I can see the result

ilayn:
You can also downscale it directly if that is the problem by the way. But it would make the table harder to read.

mdhaber:
Closed by #215

PR #215, opened 2019-07-28
Fix formatting of Refs in Table 1

FabricioS:
See #214

### DIFF
--- a/scipy-1.0/paper.tex
+++ b/scipy-1.0/paper.tex
@@ -33,6 +33,7 @@
 \newcommand{\xmark}{\ding{55}}%
 \newcommand{\code}[1]{\texttt{#1}}
 \newcommand{\fixme}[1]{\textcolor{red}{{#1}}}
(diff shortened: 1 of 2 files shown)


ilayn:
This PR 
image

Compared to previous
image

I agree it is nicer. However I am not really a fan of the slanted column names. I think typewriter font is not needed there since there is no other context and slanting seems not to help on readability.

mdhaber:
@FabricioS Thanks for this. I noticed this issue recently but didn't take the time to figure out how to fix it, so I appreciate it. 

@ilayn I think I adjusted the angle of the column names at one point. The reason I didn't go with pure vertical was that it made the table look squashed horizontally, but again, I didn't take the time to figure out how to adjust the angle and column width independently. If you can show my another option (in another PR), I'd be happy to take a look.

mdhaber:
Thanks @FabricioS!

mdhaber:
Closes #214

PR #217, opened 2019-08-12
update affiliation

VarIr:
Updated my affiliation with department info


mdhaber:
I updated this in our table. It will be added to the document before the next version is submitted.

Issue #230, opened 2019-09-06
Comment to Address: Compiled Code Coverage is Low


mdhaber:
> Page 11. The compiled coverage of 45% seems surprisingly low and perhaps should be remarked on. The maintainers also appear to have allowed remarkable amounts of untested compiled code in recent years. One could make an argument that the correctness of SciPy is much more important than the feature set as it is currently thought of as a dependable resource.

Issue #171, opened 2019-06-29
Analysis of apparent conflict between linguist and pytest-cov/gcov data


mdhaber:
`linguist`'s report that SciPy 1.0 is ~50% Python / 50% Other appears to conflict with with Figure 3, which suggests that Python makes up much less of SciPy 1.0.

This discussion began in #99. It was attributed to (reasonable) differences in the way the respective tools generated the data. I started to take a closer look.

---
I used `linguist` with nearly default<sup>1</sup> settings to calculate the number of bytes of code in each language and with the `--breakdown` option to generate a list of all the files it counted. To see whether `linguist` is excluding important files, I cross-referenced this list with the list of all files in the SciPy directory (according to `os.walk`). Here are the extensions of all files that `linguist` did not count:
```
{'', '.example', '.conf', '.ini', '.md', '.py', '.css_t', '.sh', '.npy', '.npz', '.1', '.sample', '.mat', '.png', '.dat', '.scipy', '.idx', '.x', '.svg', '.css', '.README', '.arff', '.src', '.rst', '.less', '.0rc2', '.bat', '.yml', '.nc', '.in', '.toml', '.info', '.2', '.patch', '.0', '.mp3', '.ogg', '.pack', '.js', '.wav', '.json', '.0rc1', '.pyf', '.txt', '.sav', '.0b1', '.html'}
```
 
One of the hypothesized reasons for the apparent conflict was that `linguist` tries to exclude vendored code. This could make `linguist` under-count the amount of compiled code, and thereby over-count the relative amount of Python code. I don't recognize any of the extensions above to correspond with compiled code<sup>2</sup>, so I don't think that's happening.

The other hypothesized reason for the conflict was that `linguist` counts the amount of each language in terms of bytes whereas `pytest-cov` and `gcov` report the number of lines. I used the file list from `linguist` to count the number of lines in each file to get the number of lines in each language:
```
{'Python': 282567, 'Fortran': 162212, 'C': 131534, 'Cython': 20082, 'C++': 14231, 'TeX': 1479, 'Objective-C': 152, 'MATLAB': 181, 'Shell': 107, 'Makefile': 50}
```
Python files make up 46~47% of SciPy whether you're counting bytes or lines, so that's not it either.

I wondered whether blank lines and comments could be accounting for the difference, so I separately counted blank lines and lines beginning with `#':
```
[...]

PR #236, opened 2019-09-06
Describe why code coverage is low for compiled libs

stefanv:
I made a guess here; is this the correct explanation?

Aims to address https://github.com/scipy/scipy-articles/issues/230

### DIFF
--- a/scipy-1.0/paper.tex
+++ b/scipy-1.0/paper.tex
@@ -885,7 +885,9 @@ are more than \num{13000}~unit tests in the test suite, which is written for usa
 with the \texttt{pytest}\cite{pytest} framework.
 Test coverage at the SciPy~1.0 release point (Figure~\ref{fig:coverage}) was at
 87\% for Python code according to \texttt{pytest-cov}\cite{pytest-cov} and
-45\% for compiled (C, C++, and Fortran) code according to \texttt{gcov}\cite{gcov}.
+45\% for compiled (C, C++, and Fortran) code according to
+\texttt{gcov}\cite{gcov}, the latter primarily due to the vendoring of several
+large, well-vetted Fortran libraries from repositories such as NETLIB.
 Documentation for the code is automatically built and published by
 the CircleCI service to facilitate evaluation of documentation changes /
 integrity.  Our full test suite also passes with PyPy3\cite{Bolz:2009:TMP:1565824.1565827}, a just-in-time compiled
--- a/scipy-1.0/paper.tex
+++ b/scipy-1.0/paper.tex
@@ -884,10 +884,8 @@ include code linting with \texttt{pyflakes} and \texttt{pycodestyle}. There
 are more than \num{13000}~unit tests in the test suite, which is written for usage
 with the \texttt{pytest}\cite{pytest} framework.
 Test coverage at the SciPy~1.0 release point (Figure~\ref{fig:coverage}) was at
-87\% for Python code according to \texttt{pytest-cov}\cite{pytest-cov} and
-45\% for compiled (C, C++, and Fortran) code according to
-\texttt{gcov}\cite{gcov}, the latter primarily due to the vendoring of several
-large, well-vetted Fortran libraries from repositories such as NETLIB.
+87\% for Python code according to \texttt{pytest-cov}\cite{pytest-cov}.
+Coverage for compiled (C, C++, and Fortran) code is still being implemented.
 Documentation for the code is automatically built and published by
 the CircleCI service to facilitate evaluation of documentation changes /
 integrity.  Our full test suite also passes with PyPy3\cite{Bolz:2009:TMP:1565824.1565827}, a just-in-time compiled


rgommers:
needs checking, but very likely correct

grlee77:
alternatively, is there a mechanism to exclude the vendored code from consideration in gcov? We could then report the higher number with a parenthetical (excludes vendored code).

stefanv:
> alternatively, is there a mechanism to exclude the vendored code from consideration in gcov? We could then report the higher number with a parenthetical (excludes vendored code).

Possibly, but I think this will make it hard for us to complete the edits by the end of the week, so I'd gently push back that we should see if we can get away without it.

tylerjereddy:
The pertinent discussion is in https://github.com/scipy/scipy-articles/issues/171

Matt & I did a bunch of analysis there. I think a lot of generated `.c` files are getting pick up too. We're not hiding anything--there's a docker container that can be used to reproduce / improve the cov analysis workflow if desired in the future, although it is probably just reflecting the need for easy-to-use tools in the coverage landscape that work across source / template languages.

This is still true in master--we have no active compiled code coverage reports in SciPy, just the "beta" version of coverage in Azure for pure Python, which we have to check manually (and codecov bricked for us.. and partialy for NumPy now too..).

So, in short maybe also mention questionably-relevant generated files being picked up. We can always argue this is the "harshest" / most honest coverage number.

ilayn:
I am definitely not against code coverage when it work and also not to be the party pooper but seriously who cares? It's not like academia releases very well tested code. Hardly ever if they even bother to include the code which is like never.

stefanv:
Fixed it.

tylerjereddy:
> but seriously who cares?

The reviewer -- I'm not sure that removing the number is really the best solution -- and at the 1.0 release point things were actually a bit more robust with codecov still working somewhat on PRs. I think we should just retain the original number and explain why it might be a little lower than one might expect, for now.

stefanv:
Yeah, I'm kind of torn: the original number was kind of bogus, and now we have to defend it.  What does it really say, do we know?  I'm fine with either direction.

tylerjereddy:
We have the source code that produced it, so we should know what it says. The full list of [parsed files is openly available](https://github.com/tylerjereddy/scipy-cov-track/blob/master/container_output/gcovr_details_v1.0.0.txt), and was cited in the ms I think. Even the generated and vendored code files do tend to have some % of coverage in that list.

We could regenerate the data with some filters on that stuff, but this task would likely end up falling to me. Which is probably why I'm inclined to simply cite the truth (the code) & explain why the number sounds bad but really isn't & could be "marketed" better if we used filters.

I guess we can see how much time I can come up for filtering the more marketable value if we don't want to defend the "observed truth" without filters on vendored / generated things.

ilayn:
OK apparently I jumped the gun before reading that these are all related to a recent article review. Couldn't find if it is available in the mailing list or somewhere else. 

> One could make an argument that the correctness of SciPy is much more important than the feature set as it is currently thought of as a dependable resource. 

One has to first make the argument of code coverage is a useful metric. 

> The reviewer 

I meant nobody really depends on code coverage professionally because code coverage tools are not mature enough to be dependable. We are using these numbers just as a rough idea. It is impossible to conclude correctness. Also most of the compiled code is battle-tested 3rd party libraries. So I don't think we need to defend it but mention the secondary nature of these stats.

tylerjereddy:
I've used `gcovr` exclusion filters in the analysis repo to exclude most of the Cython-generated files & a number of the vendored source paths--the [new results](https://github.com/tylerjereddy/scipy-cov-track/blob/master/container_output/gcovr_details_v1.0.0.txt#L208) show 59 % line coverage for compiled langs.

That's a fair bit better--maybe have a look at that list of files & see if anything still stands out as needing to be included / excluded. Although it took a few hours to figure out the exclusion syntax idiosyncrasies, I can probably make adjustments and get new results relatively fast now.

scipy-1.0/paper.tex
```diff
@@ -884,8 +884,8 @@ \subsubsection*{Test suite}
 are more than \num{13000}~unit tests in the test suite, which is written for usage
 with the \texttt{pytest}\cite{pytest} framework.
 Test coverage at the SciPy~1.0 release point (Figure~\ref{fig:coverage}) was at
-87\% for Python code according to \texttt{pytest-cov}\cite{pytest-cov} and
-45\% for compiled (C, C++, and Fortran) code according to \texttt{gcov}\cite{gcov}.
+87\% for Python code according to \texttt{pytest-cov}\cite{pytest-cov}.
+Coverage for compiled (C, C++, and Fortran) code is still being implemented.
```
tylerjereddy: Maybe something like this:

Coverage for compiled (C, C++, and Fortran) code was 45 %, or 59 % excluding most of the vendored (trusted external) and Cython-generated sources. Line coverage by a unit test does not guarantee correctness, and is not equivalent to formal proofing, but it can provide immediate feedback on changes in behavior.

tylerjereddy:
@mdhaber What do you think here?

mdhaber:
TLDR: Thanks for working out those filters, @tylerjereddy, but I would actually suggest keeping the 45% only and explaining it. I think the 59% number is still not really accurate because:
* `gcov` is not catching some (significant) vendored code
* the fact that I can't find some of the files in the lists suggests that they are generated (please check me on that)
* some of the files are deprecated, and so tests have probably been removed
I would prefer to give the 45% figure and explain that it is a very raw number with no corrections rather than giving the 59% figure (which is also pretty low) and explaining that it is partially, but not fully, corrected for all these factors.

Maybe something like:
> Test coverage at the SciPy~1.0 release point (Figure~\ref{fig:coverage}) was at
> 87\% for Python code according to \texttt{pytest-cov}\cite{pytest-cov}. 
> Coverage of compiled (C, C++, and Fortran) code was only 45\% to 
> \texttt{gcov}\cite{gcov}, but the compiled codebase is much more robust than
> this figure would suggest as it does not correct for the presence of reputable
> vendored code, the source of which is well-tested; generated code, 
> for which full coverage is impractical; and deprecated code, 
> which does not require unit tests.

-----

I sorted the list by files with the most uncovered lines (# of lines, not %). Here are the top 20:
```
Path - Uncovered Lines
-------------------------
*scipy/special/specfun/specfun.f   - 5415
+scipy/spatial/ckdtree.cxx - 3805
+scipy/sparse/sparsetools/csr_impl.h - 878
+scipy/sparse/sparsetools/bsr_impl.h - 766
+scipy/sparse/sparsetools/csc_impl.h - 568
scipy/interpolate/src/_fitpackmodule.c -   437
*scipy/special/cephes/jv.c - 260
scipy/signal/splinemodule.c - 211
scipy/signal/D_bspline_util.c - 257
scipy/signal/S_bspline_util.c - 257
*scipy/optimize/lbfgsb/lbfgsb.f - 246
scipy/linalg/src/lapack_deprecations/dgegv.f   - 233
scipy/linalg/src/lapack_deprecations/sgegv.f   - 233
scipy/special/specfun_wrappers.c - 220
*scipy/optimize/_trlib/trlib_tri_factor.c   - 211
scipy/linalg/src/lapack_deprecations/cgegv.f   - 208
scipy/linalg/src/lapack_deprecations/zgegv.f   - 208
+scipy/linalg/_lapack_subroutine_wrappers.f   - 194
scipy/odr/__odrpack.c - 189
```
I've `*`d the ones that appear to be vendored (manual inspection) but potentially with modifications.
Excluding these files, coverage would be at 76%.

I've `+`d the ones for which I can't find a source file with that exact name on the 1.0.x maintenance branch: https://github.com/scipy/scipy/tree/maintenance/1.0.x

https://github.com/scipy/scipy/blob/maintenance/1.0.x/scipy/interpolate/src/_fitpackmodule.c says it is "generated by setmodules.py. Do not modify it."

The `lapack_deprecations` files are deprecated, so presumably that is why they are untested.

tylerjereddy:
Feel free to push that in Matt, I'd say--that seems much more detailed & correct than just deleting the number.

stefanv:
That phrasing works for me.

mdhaber:
OK I don't think I can push to this, so please see #250.

PR #238, opened 2019-09-07
Remove unnecessary mention of CPython

### LINKED ISSUE #223: Comment to Address: Technical Details
>  Page 2. A clear distinction between CPython and Python is stressed, which might not be relevant to many audiences as alternatives like PyPI are likely not well understood. This is also an example of where the paper may delve too much into technical details.

mdhaber:
Closes #223

### DIFF
--- a/scipy-1.0/paper.tex
+++ b/scipy-1.0/paper.tex
@@ -177,10 +177,7 @@ individually, but each bringing an important piece to the puzzle.
 
 Python is an interpreted, high-level, general-purpose computer programming
 language, designed by Guido van Rossum in the late 1980s,
-with a dynamic type system and an emphasis on readability and rapid prototyping.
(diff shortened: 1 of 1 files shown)


tylerjereddy:
thanks Matt

mdhaber:
Thanks. We need to make sure this doesn't appear in the second arXiv submission though.

tylerjereddy:
I assume the backports will be up until just before the first *Nature Methods* revision is merged

mdhaber:
IIRC a publisher I'd worked with in the past didn't allow posting a manuscript that Incorporated reviewer comments. I see now that policies vary quite a bit, and Nature Methods does not seem to forbid that. Their policy is [here](https://www.nature.com/articles/nmeth.3831).

Issue #222, opened 2019-09-06
Comment to Address: Length of Background


mdhaber:
Reviewer 2:
> At a structural overview of the paper, the authors tried to both demonstrate the strengths of open-source and best practices for community projects while also attempting to detail what SciPy is. Both efforts are quite commendable and important but do make the paper a somewhat difficult read that may not satisfy either category of readers. In particular, I am not sure many readers will fully appreciate the Background section, which makes up ~1/3rd of the paper and seems somewhat misplaced as the beginning section. As is this structure will put off many readers.

>...Overall the paper is well written, but some additional attention to presenting this towards a general audience should be exercised. In particular, the “Background” section should be greatly minimized and moved to another part of the paper (or removed entirely) so that readers can be first presented with capabilities.

The journal does not think that the background section should be removed, but that it could be streamlined.

mdhaber:
I can work on this.

mdhaber:
Please keep in mind that all of this material is preserved in the first ArXiV draft, and we can even keep it in the revision to be posted if we want.

tylerjereddy:
@stefanv and @jarrodmillman may want to have a quick look at the streamlining of the background -- while perhaps not ideal, we're probably going to have to work with the journal on this one

jarrodmillman:
@mdhaber's changes look good to me.

mdhaber:
Thanks @tylerjereddy @jarrodmillman. This section started at 2000 words and we brought it down to 1505. It was a full three pages, now it's 2.5 with one of those pages being quite sparse. I think that is a solid response to the request to streamline the section.

Through this and other PRs (#234, #235, #238, #239, #240), we've also paid more "attention to presenting this towards a general audience".

I think this one is closed.

PR #243, opened 2019-09-08
Removed paragraph before SciPy Begins

mdhaber:
Partially addresses #222 

The main idea of this paragraph - that the introduction of Numeric led to the growth in the number of scientific packages, many of which were written by students - is captured in the first paragraph of the next section, SciPy Begins. The previous paragraph includes the idea that it was nice to add domain-specific functionality to a general purpose language rather than the other way around.


tylerjereddy:
Merging with Jarrod's approval

PR #251, opened 2019-09-17
WIP: Add git latexdiff script

### LINKED ISSUE #249: Underline Changes since initial submission
Once #221 and #230 are closed, we should underline the changes since initial submission as requested:

> In order to facilitate review of the revised manuscript please provide a point-by-point rebuttal and **underline any additions to the text or areas with other significant changes**. Please ensure that the revised version is as concise as possible, and that it conforms to our format requirements (see http://www.nature.com/nmeth for our Guide to Authors). We shall return to the issue of format and presentation in the event that the manuscript is accepted for publication.

tylerjereddy:
Fixes #249 but work in progress using [git-latexdiff](https://gitlab.com/git-latexdiff/git-latexdiff)

Here's an [early draft of the paper_diff.pdf](https://github.com/scipy/scipy-articles/files/3623828/paper_diff.pdf) that shows up in a temporary directory created by latexdiff when using the script in this PR, but I think I still have issues:

- [ ]  getting the `-o` flag to actually dump the PDF in a chosen working directory (I think this may be related to some compile errors)
- [ ] getting the diff to work on the bibliography (latexdiff seems to struggle with the nested style file hierarchy, maybe)

We may want to touch up the styling a bit too.

### DIFF
--- a/scipy-1.0/git_latexdiff.sh
+++ b/scipy-1.0/git_latexdiff.sh
@@ -1,0 +1,4 @@
+# use git-latexdiff to provide a highlighted PDF with changes
+# requested for Nature Methods
+
+git latexdiff --main paper.tex --output paper_diff.pdf 6ff8517237 master


tylerjereddy:
I've not had much luck after a few hours of messing around to get the diff to work on the reference list as well. Perhaps those recommending the git-latexdiff/ similar tools (@ev-br / @jarrodmillman ) want to take a look?

The body of the text has an ok diff though.

ev-br:
Does this help?https://tex.stackexchange.com/questions/172094/latexdiff-doesnt-mark-changes-in-bibliography-bibtex

tylerjereddy:
I saw that, but I don't really get it, since there is support for reference diffing in the help for git-latexdiff. Maybe it is broken for more complex projects. If we need to string together a script with several commands feel free to give that a go!

ev-br:
(am on the phone ATM, sorry). Can you try `--bbl` option? 
https://gitlab.com/git-latexdiff/git-latexdiff/issues/14

tylerjereddy:
I tried --bbl, bibtex, and various other options listed in the help related to bibtex. I also had to copy our specific style files to an appropriate system path, possibly because git-latexdiff uses a tempdir.

I think this may take some effort to resolve. Perhaps someone will just run a separate command for the references when we are ready to resubmit, which is hopefully soon.

jarrodmillman:
The effort to track this down is not worth it.  Reviewers will only care about the differences to the text body.  I would just mention that the changes to the bibliography aren't color coded due to technical issues and leave it at that.  However, I didn't see the actual reviews.  If one of the reviewers specifically asked for a reference to be added, removed, or modified, you should just mention that the modification where made (or not and why) in the "rebuttal" letter.

tylerjereddy:
> The effort to track this down is not worth it.

Right, that's what I thought--this actually seems like a pretty nasty thing.

mdhaber:
I don't think we got any comments about references.
@tylerjereddy Does Table 1 in the latexdiff output still extend beyond the edge of the page?
I will manually indicate changes to the bibliography if need be.

tylerjereddy:
> Does Table 1 in the latexdiff output still extend beyond the edge of the page?

Yes, I think so. I didn't really try to adjust the formatting, I just tried to get the references working so I think what you see in this PR is still what we will end up with barring any further modifications.

> do you have the .bbl file (or commit number) of the version originally submitted to Nature Methods? I can use git to find the diff compared to today.

We have a git hash used in this PR for the comparison--that's an estimate based on submission date, but should be pretty close & sufficient to regenerate the bbl?

> I will manually indicate changes to the bibliography if need be.

Well, I guess I can do that... we need reliable tools for this stuff eventually. That's why I originally suggested using a highlight markup when we made the changes... Off the top of my head, there's the Open Hub scitation, the new citation to the Nature News article about CI, and a citation from a test-driven development textbook added.

Still we are currently 2 to 1 for core devs in favor of basically claiming the diff doesn't work so well for the references & those ones I mention are the main ones I think so just mentioning those may be sufficient.

tylerjereddy:
>  Chances are that most authors (who are not necessarily TeX-savvy) will manually indicate changes, and there are no technical hurdles for doing that.

Sure there are, I doubt the manual approach is completely immune to forgetting to highlight changes, especially on big multi-author papers with Microsoft Word & track / accept changes.

mdhaber:
I don't think it's actually important; I would just prefer not to appear lazy to the reviewers. Anyway, I have the diffs; it's just a matter of adding lines in the final pdf. I'm happy to do it, unless there is opposition to it being done manually.

tylerjereddy:
The key here was to add `--append-safecmd=cmd1,cmd2` where the `cmd` values are appropriate bibtex commands, when using `latexdiff`. Unfortunately, I spent a few hours trying to do it manually before I discovered that.

see: https://github.com/tylerjereddy/nmeth_resubmit

PR #253, opened 2019-09-18
MAINT: move CI definition

### LINKED ISSUE #252: Move the CI definition
See my comment here: https://github.com/scipy/scipy-articles/pull/239#issuecomment-532498237

In trying to help with the rebuttal letter I noticed that the CI definition happens after the point where the reviewer requests it instead of before.

Maybe we can shift it a bit?

tylerjereddy:
Fixes #252 to move the CI definition prior to the first mention of CI in a Figure, as per reviewer comment.

While the CI definition may be slightly out of place at that point of the article, I'd say checking off the box to satisfy the reviewer comment is more important that some kind of ideal / flawless flow to the paper with all testing content confined to the testing section.

### DIFF
--- a/scipy-1.0/paper.tex
+++ b/scipy-1.0/paper.tex
@@ -519,7 +519,13 @@ and milestones in its history are highlighted in Figure~\ref{fig:timeline}.
 \centering
 \includegraphics[width=0.95\textwidth]{static/scipy_timeline}
 \caption{Major milestones from SciPy's initial release in 2001 to
-the release of SciPy 1.0 in 2017.}
+the release of SciPy 1.0 in 2017. Note that SciKits and GitHub 
+have been introduced in the ``Background'' section;  more 
+information about Cython and SciPy subpackages (e.g. 
+\texttt{scipy.sparse}) is available in ``Architecture 
+and implementation choices'', BLAS/LAPACK support is 
+detailed in ``Key technical improvements'', and continuous
+integration is discussed in ``Test and benchmark suite''.}
 \label{fig:timeline}
 \end{figure}
 


scipy-1.0/paper.tex
```diff
@@ -503,6 +503,12 @@ \subsection*{SciPy Today}
 \num{600000}~lines of code organized in 16~subpackages. The development
 team and community currently interact and operate primarily
 on GitHub, an online version control and task management platform.
+According to the practice of continuous integration\cite{silver2017collaborative},
+all proposed contributions to SciPy via GitHub are temporarily integrated with the
+master branch of the library
+before the test suite is run, and all tests must pass before the contribution
```
mdhaber: If the reader is not familiar with "continuous integration", they might not be familiar with "test suite".

mdhaber: Instead of moving that part, how about adding a bit of detail to the figure caption. After all, if the reader is not familiar with continuous integration, they may not be familiar with a bunch of other things mentioned in the figure that haven't been described yet, like Cython, BLAS/LAPACK, and GitHub. I'll suggest some text.

mdhaber: How about leaving the definitions where they make sense and referring the reader to them here? I don't think we need to do _exactly_ what the reviewer says; I think this would adequately address the comment without creating new problems.

Caption of Figure 2:
```
Major milestones from SciPy's initial release in 2001 to the release of SciPy~1.0 in 2017. Note that SciKits and GitHub have been introduced in the ``Background'' section; more information about Cython and SciPy subpackages (e.g. \texttt{scipy.sparse}) is available in ``Architecture and implementation choices'', BLAS/LAPACK support is detailed in ``Key technical improvements'', and continuous integration is discussed in ``Test and benchmark suite''.
```

I wish we could use section/subsection references....

tylerjereddy: I'd rather just do what I did, but fair enough. I'll do this--can someone get the latexdiff thing working with references + bibtex in the meantime?

tylerjereddy:
Alright, I've revised & force-pushed to match your request. We'll need a short rebuttal modification/ addition once this goes in.

Issue #263, opened 2019-09-26
Editorial Revision: Artwork Guidelines


tylerjereddy:
Attachment from *Nature Methods* Editorial team wrt formatting of images:

[NMETH-P39410A_Figure_Guidelines_1569514283_1.pdf](https://github.com/scipy/scipy-articles/files/3658908/NMETH-P39410A_Figure_Guidelines_1569514283_1.pdf)

ilayn:
@mdhaber Should we assign ourselves to this ?

mdhaber:
I'll take a more detailed look and let you know.

tylerjereddy:
@mdhaber it looks like Figure 4 *might* already be in "vector" format (they accept PDF); I suppose it may need a slight resizing to match their guidelines for two-column width & height limits & possibly a font check to match their requirements

tylerjereddy:
I imagine for Figure 3, it shouldn't be too hard to split that tikz code out of the manuscript & produce a standalone vector PDF as per their requirements.

ilayn:
Figure 3 is already in vector format but sure we can separate the file out to a PDF. I think For fig 2 and 4, Matt has the inkscape files (and also TikZ version in #179). 

I think we need to change the caption of Fig. 1 to maybe a quote. Feels strange to have it as a figure.

tylerjereddy:
> I think we need to change the caption of Fig. 1 to maybe a quote. Feels strange to have it as a figure.

Yeah, I think the specific request was to change it to a "Box" that wouldn't count against the word count total.

mdhaber:
@ilayn I just created #263 for figures 2 and 4. If you could take care of 3 (please use Arial instead of Helvetica), I'd appreciate it. For consistency, could you use two of the colors from Figure 4?

It sounds like 1 needs to be changed in MS word?

mdhaber:
I'd suggest that we don't change these in the LaTeX document; I would prefer the existing font for arXiv version 2.0.

mdhaber:
@ilayn will you be able to provide the code coverage figure?

ilayn:
@mdhaber I am a bit lost in the requirement between the Arxiv and the journal submission. Which one should I convert ? 

The fonts are taken from the surrounding TeX document hence I am not selecting any fonts.

mdhaber:
[...]

PR #265, opened 2019-09-29
Prepared timeline and ASV figures for Nature Methods

mdhaber:
Partially addresses #263 

- Set width to 180 mm
- Converted text (except variable symbols) to Arial or Consolas 7pt
- Exported as PDF

Files were originally prepared in Inkscape, which does not natively produce CMYK files, so it is impossible to satisfy both requirements:

- All other content (including Perspectives, Progress Articles and Review Articles) should be supplied in CMYK colour mode.
- All line art, graphs, charts and schematics should be saved/exported directly from the original application and file in which they were generated.

So the files are in RGB, and I suggest that we submit them that way, explaining the dilemma, and offering that the journal auto-converts to CMYK, it will not affect the content that the figures convey.


mdhaber:
Closes #268

stefanv:
@mdhaber Could you check the text alignment for 2011?

You can apparently use [Krita](https://krita.org) or an [inkscape extension](https://wiki.inkscape.org/wiki/index.php/ExportPDFCMYK) to get CMYK, although I've never done this myself.

ilayn:
Inkscape should be capable of choosing a color model for this. But in any case I think we can remove all the logos since now it is half text half logo'd. Or we can use the TeX version where you can change the color model.

mdhaber:
@stefanv is this better? Thanks for the tip about the extension. I knew I could convert to CMYK using some post-processing but the instructions also (surprisingly) specified that the PDF was to be produced with the tool used to create the file, so I chose which of the two conflicting instructions to follow.

@ilayn all the logos were removed before the recent commits. What is the half-logo'd version?

The journal doesn't seem to be complaining about the color model, so I don't think we need to worry about it. I'm fine with going with the TeX version at this point if it turns out to be important.

mdhaber:
@tylerjereddy I tweaked an alignment in the timeline per @stefanv's suggestion. Could you send them the latest version?

Shall we merge this PR so the files are in the repo?

ilayn:
I meant all logos including numpy etc. a simple timeline can also cut down some vertical space.

tylerjereddy:
I think the paper was formally accepted before the revised version of the Figure was made available. Once Matt stripped out the copyrighted logos, as the Editorial Team requested, I sent that version.

Shifting the text "Development moves to GitHub" 1 cm to the right doesn't seem worth the churn IMO.

 I've already spent a lot of time dealing with stuff related to the process, so unless this was a pretty big blemish I'd rather not do more submission stuff.

It doesn't look like a big deal to me. That said, Matt or Ralf are likely more than welcome to touch base with our Editorial contact with the updated Figure since they are both corresponding authors as well.

ilayn:
I don't have any strong opinions but **if** we are going to touch the figure we better remove all logos. Otherwise I am fine with it.

tylerjereddy:
Removal of **all** logos was a requirement for publication--this is the version sent to the journal for publishing:

image

ilayn:
Hmm I don't get it. In the artifacts of CircleCI I still see the logos. Maybe that's why we are talking past each other.

rgommers:
That figure looks better than the original I think, very clean!

mdhaber:
@ilayn again, there are no logos in the version of the figure accepted by Nature Methods. See `scipy_timeline_NM2.pdf`, which is part of this PR. It is just text and lines. We're just talking about a text alignment issue now.

@tylerjereddy It is not a big deal. I will see if they will change it without trouble, and if not, I'll drop it.

ilayn:
Well again, there are logos present in the current CircleCI artifacts paper.pdf 

Do you mind if I just get out f this? Since I don't have any valuable input.

ilayn (approved):

mdhaber:
@ilayn Sorry, I didn't see the comment about CircleCI before. The CircleCI artifacts are not used these days because Nature Methods required that we submit the article as a Word document, so we haven't been updating the TeX document.

stefanv:
Thanks for the tweak @mdhaber, that looks great.  I thought updating it would simply require sending an email, but if it's any hassle at all don't worry about it.

PR #285, opened 2026-08-17
clean-up some comments from the template

j-bowhay:
cleans up some guff making the document slightly easier to navigate


j-bowhay:
self merged as this has no impact on the actual content

PR #280, opened 2026-08-08
Add section on Fortran rewrite

j-bowhay:
closes #278

A first draft at trying to discuss the benefits we have seen from rewriting while trying to avoid a "get rid of Fortran" and "Fortran free" sentiment. Very open to suggestions:)

PR #292, opened 2026-09-11
Add plot showing how SLOC by language has changed over time

j-bowhay:
@lucascolley prompted by our prior discussion (https://github.com/scipy/scipy-articles/pull/280#discussion_r3925097807), I was think about how to better convey how SciPy's language choices have changed. I concluded that I don't really like just stating percentages as it is pretty dry to read and doesn't really paint the full picture. I decided it would be better to include a plot of how the language choices have changed since SciPy 1.0.
image

I decided we should include to `subprojects` directory otherwise it looks like code has just disappeared into thin air over time. I know we could make counter arguments like "we don't use all of Boost" and "we are not responsible for maintaining all this code" but I think this best illustrates what we want to show.

WDYT?


ilayn:
stacked area plots are better when you are trying to breakdown an aggregate, instead of the implied racing nature of line plots. it can also show the disappearing of F77 while total sloc is increasing. we can also mark boost propack highs and unuran additions to make the bumps more meaningful

lucascolley:
> stacked area plots are better when you are trying to breakdown an aggregate, instead of the implied racing nature of line plots. it can also show the disappearing of F77 while total sloc is increasing. we can also mark boost propack highs and unuran additions to make the bumps more meaningful

yeah this sounds like it could be very cool with a bit of effort

ilayn:
if you have the data I can cook up the pgfplots version https://tikz.dev/pgfplots/reference-2dplots#sec-4.5.10.1 It's not much work

j-bowhay:
@ilayn @lucascolley is this what you had in mind?
image
The disadvantage is that you can no longer use a logarithmic y scale because it distorts the proportion of each language.

Let me know which you'd prefer!

The massive increase and then decrease in C++ is because originally we pulled in the entirety of Boost rather than just Boost.maths

ilayn:
i think the entirety of boost or not won't differ much since we did not use it or linked to it. So we can cap at boost.math size regardless. Then the visual would me more sense in linear scale. especially if we put c and c++ at the bottom.

j-bowhay:
Thanks @ilayn that looks way clearer
image

> we can also mark boost propack highs and unuran additions to make the bumps more meaningful

Any clever suggestions on marking these without making the figure too crowded?

ilayn:
Unfortunately, I only know how to do it in latex/tikz 😕

j-bowhay:
> Unfortunately, I only know how to do it in latex/tikz 😕

Do you have an example you can point to? It's more that I am struggling to visualise what you mean

ilayn:
image


overlayed an idea in TikZ with very little attention to styling and fonts. yo can also use brackets if it is spread over multiple versions for slow increases instead of arrows

j-bowhay:
Ok thanks I'll have a play with that

ev-br:
Quick question. What does this figure illustrate, the code that gets installed or the code SciPy maintains. I'd think the latter is what we want to show? So that "Fortran is gone" is visually prominent, while the the figure is dominated by HIGHs and boost.math instead.

j-bowhay:
> Quick question. What does this figure illustrate, the code that gets installed or the code SciPy maintains. I'd think the latter is what we want to show? So that "Fortran is gone" is visually prominent, while the the figure is dominated by HIGHs and boost.math instead.

Good question, I went back and forth with myself on this one a LOT. In the end I decided on everything in the `scipy` and `subprojects` directories, this isn't perfect but I decided on it because:

- The audience of the paper is user facing rather than dev facing so I wanted to give an indication of how the overall complexity of SciPy has changed not just the bits we are directly responsible for maintaining.
- Because in older versions we used to vendor things in a random assortment of ways its really difficult to get a consistent count across versions if you want to exclude things as they keep moving about. Plus it is not always clear what we maintain, what we sort of maintain, and what we don't.

WDYT @ev-br?

ev-br:
First and foremost: I'm happy with either choice you make :-). If and only if we're going into details, my slight preference would be to avoid the message "scipy is now written in C++", and make the Fortran removal mode prominent: the figure as is only shows a narrow green band which starts dwindling around scipy 1.16. So it was not a big deal and keeps not being a big deal, it looks like. IOW, in the figure as drawn, adding boost.math is a large event, while all Ilhan's Fortran rewrites pale in comparison. And the transition to `xsf` is not visible at all. OTOH, all these concerns can be addressed in text.

lucascolley:
I have a preference for including — maybe alone, or maybe alongside this plot — a plot restricted to 'code that SciPy maintainers are responsible for maintaining'. I know that isn't the easiest thing to get right (especially for older versions), but I think we can take the liberty to say we were basically responsible for everything in the codebase back in the day, then exclude the biggest SLOC contributors (HiGHs, boost) as they were introduced. That should make the Fortran removal more visually striking.

It would also be cool to mark, on that plot, when xsf was moved out into a separate repository.

j-bowhay:
> I have a preference for including — maybe alone, or maybe alongside this plot — a plot restricted to 'code that SciPy maintainers are responsible for maintaining'. I know that isn't the easiest thing to get right (especially for older versions), but I think we can take the liberty to say we were basically responsible for everything in the codebase back in the day, then exclude the biggest SLOC contributors (HiGHs, boost) as they were introduced. That should make the Fortran removal more visually striking.

Do you think we could convey that information in a single plot by using solid colour to represent code we maintain and either hatching or transparency for code we aren't?

My concern is that two separate plots is a lot of real estate to give to the topic. My guiding principal is that the paper should be a 10-20 page exposition of the changes since SciPy 1.0 that _users_ should care the most about, rather than being an exhaustive treatise of SciPy development. As a result, I think we will need to prioritise some details over others. However, maybe I am under estimating the importance of the topic and the space for two figures is deserved. Maybe there's a clever way to display all this information compactly. (Or maybe it is my subconscious making me stubborn at the thought of following Boost and other libraries around the SciPy source tree!!) WDYT?

> It would also be cool to mark, on that plot, when xsf was moved out into a separate repository.

Just for clarification, in your proposal, which plot would contain `xsf`?

lucascolley:
> > I have a preference for including — maybe alone, or maybe alongside this plot — a plot restricted to 'code that SciPy maintainers are responsible for maintaining'. I know that isn't the easiest thing to get right (especially for older versions), but I think we can take the liberty to say we were basically responsible for everything in the codebase back in the day, then exclude the biggest SLOC contributors (HiGHs, boost) as they were introduced. That should make the Fortran removal more visually striking.
> 
> Do you think we could convey that information in a single plot by using solid colour to represent code we maintain and either hatching or transparency for code we aren't?
> 
> My concern is that two separate plots is a lot of real estate to give to the topic. My guiding principal is that the paper should be a 10-20 page exposition of the changes since SciPy 1.0 that _users_ should care the most about, rather than being an exhaustive treatise of SciPy development. As a result, I think we will need to prioritise some details over others. However, maybe I am under estimating the importance of the topic and the space for two figures is deserved. Maybe there's a clever way to display all this information compactly. (Or maybe it is my subconscious making me stubborn at the thought of following Boost and other libraries around the SciPy source tree!!) WDYT?

I think a single plot which excludes Boost and HiGHs SLOC would probably be sufficient, with a sentence or two explaining that many calculations in these areas are now offloaded to dedicated C++ projects with a more focused scope. Even from a user-perspective, I think the argument can be made that it is more important to convey 'SciPy 2.0 now has less {insert mysterious under-maintained compiled code here} and more Python', than 'SciPy now uses parts of some massive C++ libraries'. The latter mainly matters _because_ it has enabled the former, insofar as the code users are relying on has become better maintained.

> > It would also be cool to mark, on that plot, when xsf was moved out into a separate repository.
> 
> Just for clarification, in your proposal, which plot would contain `xsf`?

I would include it in the 'maintained by SciPy plot', but perhaps in hatching for the parts moved out to a separate repository. This is a minor detail that doesn't matter too much, though.

j-bowhay:
> > > I have a preference for including — maybe alone, or maybe alongside this plot — a plot restricted to 'code that SciPy maintainers are responsible for maintaining'. I know that isn't the easiest thing to get right (especially for older versions), but I think we can take the liberty to say we were basically responsible for everything in the codebase back in the day, then exclude the biggest SLOC contributors (HiGHs, boost) as they were introduced. That should make the Fortran removal more visually striking.
> > 
> > 
> > Do you think we could convey that information in a single plot by using solid colour to represent code we maintain and either hatching or transparency for code we aren't?
> > My concern is that two separate plots is a lot of real estate to give to the topic. My guiding principal is that the paper should be a 10-20 page exposition of the changes since SciPy 1.0 that _users_ should care the most about, rather than being an exhaustive treatise of SciPy development. As a result, I think we will need to prioritise some details over others. However, maybe I am under estimating the importance of the topic and the space for two figures is deserved. Maybe there's a clever way to display all this information compactly. (Or maybe it is my subconscious making me stubborn at the thought of following Boost and other libraries around the SciPy source tree!!) WDYT?
> 
> I think a single plot which excludes Boost and HiGHs SLOC would probably be sufficient, with a sentence or two explaining that many calculations in these areas are now offloaded to dedicated C++ projects with a more focused scope. Even from a user-perspective, I think the argument can be made that it is more important to convey 'SciPy 2.0 now has less {insert mysterious under-maintained compiled code here} and more Python', than 'SciPy now uses parts of some massive C++ libraries'. The latter mainly matters _because_ it has enabled the former, insofar as the code users are relying on has become better maintained.

To play devil's advocate, why exclude Boost and HiGHs but not Qhull etc etc? I ask as my original idea was similar to this but I ended up talking myself out of it because the distinction felt slightly arbitrary.

lucascolley:
> To play devil's advocate, why exclude Boost and HiGHs but not Qhull etc etc? I ask as my original idea was similar to this but I ended up talking myself out of it because the distinction felt slightly arbitrary.

Boost and HiGHs would be a good starting point, but we can put more effort in to be more precise if deemed worthwhile. It might make sense to exclude Qhull from the start too, yeah (hard to say how active the upstream is, really). Hopefully https://github.com/scipy/scipy/issues/21232 can be a helpful reference for this.

j-bowhay:
Ok, I'll have a go a making a version that contains only code that SciPy maintains and we can decide from there.

j-bowhay:
@lucascolley @ev-br after an afternoon of archeology I now have the plot which includes ONLY the code SciPy maintains:
image

I have put this in the main body of the text. For completeness, I have included the full plot in an appendix.

[paper.pdf](https://github.com/user-attachments/files/32421254/paper.pdf)


Let me know what you think!

ev-br:
Thanks Jake! IMO this one is much more illustrative, and it demonstrates the evolution of the SciPy codebase much better. The effort you put into it is really well spent IMO!

ev-br:
While details of the work to produce the figure is still fresh, could you share somewhere your scripts, notes or prompts, whichever ways you used to collect the data for the plot?
As any interesting result, it generates many follow-up questions; while we as authors don't have to have a detailed answer to all possible questions from the readers :-), we can totally expect questions from referees which will be much easier to respond to with your notes. 
Related, does the target journal accept Supplemental Materials? Details can go in there, for instance.

lucascolley:
> While details of the work to produce the figure is still fresh, could you share somewhere your scripts, notes or prompts, whichever ways you used to collect the data for the plot?
> As any interesting result, it generates many follow-up questions; while we as authors don't have to have a detailed answer to all possible questions from the readers :-), we can totally expect questions from referees which will be much easier to respond to with your notes.

the idea is that it's a single line, `pixi run plot-sloc` :) I'm trying to reproduce now

lucascolley:
To reproduce on osx-arm64, I did:
- Install `BasicTex.pkg` from https://www.tug.org/mactex/morepackages.html
- `$env.PATH = ($env.PATH | append /usr/local/texlive/2026basic/bin/universal-darwin)`
- `sudo tlmgr option repository https://mirror.ctan.org/systems/texlive/tlnet`
- `sudo tlmgr install type1cm cm-super dvipng libertine`
- `pixi run plot-soc`

Unfortunately, the LaTeX toolchain on conda-forge has been in need of some TLC for at least a decade now (e.g. https://github.com/conda-forge/texlive-core-feedstock/issues/19), otherwise we could remove the first 4 steps too.

scipy-2.0/scripts/sloc_plot.py
```diff
@@ -0,0 +1,203 @@
...
+    labels=["Python", "Cython", "Fortran 77", "C", "C++"],
+)
+ax.set_xlabel("SciPy Version")
+ax.set_ylabel("SLOC")
+ax.set_xticks(range(len(sloc_data["Version"])))
+ax.set_xlim(0, len(sloc_data["Version"]) - 1)
+ax.set_xticklabels(s.rstrip(".x") for s in sloc_data["Version"])
+fig.legend(ncols=5, loc="outside upper center")
+plt.xticks(rotation=45)
+plt.show()
```
lucascolley, line 165: can we have the script write the output to `src/figures` directly?

j-bowhay: I have to admit I prefer manual saving so that any updates are an explicit choice but happy to yield if others feel strongly

lucascolley: Good point, it should be explicit, but maybe we can add a flag for it? I was hoping to be able to see whether or not the image changed due to changes in the script, but maybe that's not possible with a (timestamped?) PDF anyway

j-bowhay: Done in latest commit

j-bowhay:
> To reproduce on osx-arm64, I did:
> 
> * Install `BasicTex.pkg` from https://www.tug.org/mactex/morepackages.html
> * `$env.PATH = ($env.PATH | append /usr/local/texlive/2026basic/bin/universal-darwin)`
> * `sudo tlmgr option repository https://mirror.ctan.org/systems/texlive/tlnet`
> * `sudo tlmgr install type1cm cm-super dvipng libertine`
> * `pixi run plot-soc`
> 
> Unfortunately, the LaTeX toolchain on conda-forge has been in need of some TLC for at least a decade now (e.g. [conda-forge/texlive-core-feedstock#19](https://github.com/conda-forge/texlive-core-feedstock/issues/19)), otherwise we could remove the first 4 steps too.

@lucascolley in the latest commit I have (hopefully) set `matplotlib` to use `tectonic`. Does this resolve these issues?

ilayn:
You can also offload drawing to TeX side instead installing a lot of dependencies. Ran the script on an annoying Surface Pro which runs everything wrong on windows 10. So Cython numbers came out 0, and matplotlib failed but managed to get the csv out from dataframe named `slocdata.csv`

```csv
Version,Python,C,C++,Fortran 77,Cython
1.0.x,218366,34931,4642,80401,0
1.1.x,229942,35816,4610,79718,0
1.2.x,247677,36466,4610,79402,0
1.3.x,252019,36343,4181,79415,0
1.4.x,264808,35770,4370,75254,0
1.5.x,274065,35627,4364,75263,0
1.6.x,282052,36217,4355,74265,0
1.7.x,298285,37036,8429,74266,0
1.8.x,314090,35737,10246,105904,0
1.9.x,326853,36074,10242,104823,0
1.10.x,341212,36507,10232,104822,0
1.11.x,356882,36375,10308,104463,0
1.12.x,374278,46264,13519,103743,0
1.13.x,378518,60067,14390,83708,0
1.14.x,381003,66847,19056,83708,0
1.15.x,410518,78022,27316,70735,0
1.16.x,421876,83222,33832,67933,0
1.17.x,437519,102658,36358,19308,0
1.18.x,451084,109211,40040,8390,0
2.0.x,450898,109907,64568,0,0
```

Then you can generate whatever image you want in the style of the article with matching fonts etc. You can also externalize or add the resulting pdf and so on. I manually placed the legend etc. to make things visible. LLMs seem to know pgfplots and TiKz now.

```latex
\documentclass{standalone}
\usepackage{pgfplots}
\pgfplotsset{compat=1.18}

\begin{document}

\pgfplotstableread[col sep=comma]{slocdata.csv}\loadedtable

\begin{tikzpicture}
    \begin{axis}[width=10cm, height=4cm,
    ylabel=SLOC,
    xlabel=SciPy Version,
        stack plots=y,
        area style,
        enlarge x limits=false,
        xtick=data,
        xticklabel style={rotate=45},
        yticklabel style={/pgf/number format/.cd,fixed},
        scaled y ticks=false,
        symbolic x coords = {1.0.x,1.1.x,1.2.x,1.3.x,1.4.x,1.5.x,1.6.x,1.7.x,1.8.x,1.9.x,1.10.x,1.11.x,1.12.x,1.13.x,1.14.x,1.15.x,1.16.x,1.17.x,1.18.x,2.0.x},
        xticklabels={1.0,1.1,1.2,1.3,1.4,1.5,1.6,1.7,1.8,1.9,1.10,1.11,1.12,1.13,1.14,1.15,1.16,1.17,1.18,2.0},
        legend style={
            area legend,
            at={(0.5,-0.75)},
            anchor=north,
            legend columns=-1,
        },
    ]
        \addplot table [x=Version,y=Python] from \loadedtable \closedcycle;
        \addplot table [x=Version,y=Cython] from \loadedtable \closedcycle;
        \addplot table [x=Version,y=Fortran 77] from \loadedtable \closedcycle;
        \addplot table [x=Version,y=C] from \loadedtable \closedcycle;
        \addplot table [x=Version,y=C++] from \loadedtable \closedcycle;
        \legend{Python,Cython,Fortran 77,C,C++}
        
        
    \end{axis}
\end{tikzpicture}

\end{document}

```

image

lucascolley:
should we follow GitHub in making C++ Red, Fortran Purple, and C grey?

ilayn:
I am colorblind so I'll leave that to you folks 😅 pgfplots support all kinds of colormaps or custom colors in the TiKz manual https://tikz.dev/tikz-actions#pgf./tikz/color and xcolor package. I could not run the article class on thisntablet so it shows as default serif TeX font in the screenshot. It just needs the CSV data.

lucascolley:
> @lucascolley in the latest commit I have (hopefully) set `matplotlib` to use `tectonic`. Does this resolve these issues?

works like a charm!!

lucascolley:
I wonder whether we can teach https://github.com/XAMPPRocky/tokei to recognise Pythran

j-bowhay:
@lucascolley I tried the Github colors but it don't like how similar Python and Fortran look so am leaning towards sticking with the original. WDYT?
image

lucascolley:
I'd suggest going for *similar* colours to what GitHub uses, but we can change shades and so on to make it look nicer.

j-bowhay:
Ok, this is ready to go from my side
[paper.pdf](https://github.com/user-attachments/files/32488172/paper.pdf)

lucascolley (approved):
LGTM, great work Jake!

I think it's worth opening an issue about Pythran not being included (or mentioned) in the SLOC figures. Probably not the highest priority thing to be working on, though.

j-bowhay:
Thanks all! Although this took quite a lot of work it ended up much better than my original idea

